Issue #93, opened 2013-06-27
Library autodetection doesn't work when one library uses another library


matthijskooijman:
I'm using the ShiftOutX library, which uses the SPI library. Arduino-Makefile fails to detect this case, resulting in a SPI.h not found error message.

pglu:
Hello, is this issue still unsolved?

xxxajk:
I'm happy to see someone else has forked the project!
That said...
add this to the top of your .ino, and it will work IDE or not: `#include <Arduino.h>`

After you do that, it *will* detect everything, At least it does in my trunk, which this fork is based on...

pglu:
Hi, thanks for your quick reply, I am also happy to be able to use your project. However I'm still struggling with some libraries, but will check out including Arduino.h. The reason may be, that on my Raspi IDE 1.0.5 is instalIed don't know how to update it.

sej7278:
No need to include Arduino.h, read the source, that's already done.

And this isn't a fork of anything from xxxajk, don't feed the troll

xxxajk:
While this isn't a fork of my effort, both have a common root, which IIRC is pre-github days.

sej7278:
Yup, first version of this we have history of this was [2010](https://github.com/sudar/Arduino-Makefile/blob/master/HISTORY.md#2010-05-21-version-03)

PR #238, opened 2014-08-14
Add Support for Library Auto-detection

ladislas:
## Note

This a work in progress, made available for testing purposes and collaboration.

Here is the first pull request regarding library auto-detection when one library uses another library, as discussed in https://github.com/sudar/Arduino-Makefile/issues/93
## Goal

Right now we have to `#include` all the libraries used by the sketch and by any other library by hand. This can become quickly cumbersome when you use a lot a libraries...

With library auto-detection, no need to do that anymore.
## Current status

_It works_. I've added an `AUTO_LIB` variable that you can set in your `Makefile` to:
- `1` - to activate auto-detection _or_ use the Makefile as usual (with all libraries in the same sketch)
- `0` - to use the Makefile as usual

It works by listing every header file in your `USER_LIB_PATH` folder and including them while compiling.

Even though it's working, the solution is not very elegant.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -736,7 +736,7 @@ ifeq ($(strip $(CHK_SOURCES)),)
                 $(call show_config_info,No .pde or .ino files found. If you are compiling .c or .cpp files then you need to explicitly include Arduino header files)
             else
                 #TODO: Support more than one file. https://github.com/sudar/Arduino-Makefile/issues/49
-                $(error Need exactly one .pde or .ino file. This makefile doesn't support multiple .ino/.pde files yet)
+                $(error Need exactly one .pde or .ino file. This makefile doesn\'t support multiple .ino/.pde files yet)
             endif
         endif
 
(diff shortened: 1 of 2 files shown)


ladislas:
Hi there! I think you can delete the auto-lib branch. I'm keeping my fork but won't have time to merge it back here for the moment :)

lluiscampos:
Hi @sudar and @ladislas,

Extending the library auto-detection for library files is a very interesting feature. What is the remaining work to bring `auto-lib` branch to `master`? I am willing to work on it but I need some guidance.

> I haven't had the time to make it work if the name differs,

I had a look at the script in `auto-lib` branch and I could take care of this.

> but I already see a problem: if two .h files have the same name in different folders, it will not work...

Can you give some more detail on this?

Why not?

I can see that having two header files with the same name in different libraries can lead to problems, as we'll rely on ordering for the compiler choosing the right one. But as long as we keep this feature optional (iow, following Arduino IDE features) is up to the users to risk such problems. We should document this limitation and risk.

Generally speaking it is not a good practice to override header files, so your libraries shouldn't share filenames.

Thanks!

ladislas:
@lluiscampos you should open a new issue to discuss this, it's not ideal to use a PR closed a long time ago.

PR #310, opened 2015-01-01
Change CXX_NAME to CXX as per the EmacsWiki for Flymake instructions.

sej7278:
Makes it a bit more standard too.


sudar:
Thanks. Just merged it.

PR #311, opened 2015-01-07
Provide option for less verbose `make` output.

trombonehero:
Allow autoconfiguration output to be silenced by setting the ARDUINO_QUIET_MAKE variable.

### DIFF
--- a/Common.mk
+++ b/Common.mk
@@ -16,6 +16,7 @@ runscript_if_exists =                                                          \
 # the number of bytes indicated by the second argument.
 space_pad_to = $(shell echo $(1) "                                                      " | head -c$(2))
 
+ifndef ARDUINO_QUIET_MAKE
 # Call with some text, and a prefix tag if desired (like [AUTODETECTED]),
 show_config_info = $(call arduino_output,- $(call space_pad_to,$(2),20) $(1))
 
@@ -28,6 +29,7 @@ show_separator = $(call arduino_output,-------------------------)
 
 $(call show_separator)
 $(call arduino_output,Arduino.mk Configuration:)
+endif   # !ARDUINO_QUIET_MAKE
 
 ########################################################################
 #


sudar:
We already have a variable called [`ARDUINO_QUIET`](https://github.com/sudar/Arduino-Makefile/blob/master/arduino-mk-vars.md#arduino_quiet) which turns off verbose output.

Does your implementation provide any further advantage?

sej7278:
this does look like it offers identical functionality to ```make ARDUINO_QUIET=1``` in that it stops the config output being displayed...

sudar:
@sej7278 Thanks.

@trombonehero I think the name `ARDUINO_QUIET_MAKE` is slightly confusing since we already have a variable called `ARDUINO_QUIET`.

Can you rename it to `DISABLE_CONFIG_OUTPUT` and add a note about it in the [arduino-mk-vars.md](https://github.com/sudar/Arduino-Makefile/blob/master/arduino-mk-vars.md) file as well?

sej7278:
hang on, why merge this when it does something we already have?

trombonehero:
Hi,

Sorry for tuning out from my own pull request yesterday, but the reason for
adding this behaviour is that I didn't spot the existing one. Now that
you've pointed out the existing thing, I don't see a need to merge my pull
request. Sorry for the bother.

Cheers,

Jon

sudar:
> hang on, why merge this when it does something we already have?

My bad.

> this does look like it offers identical functionality ...

For some reason I read this as 

> this does look like it offers **additional** functionality ...

and assumed that `ARDUINO_QUIET` doesn't allow you to hide config output and this pull request allowed you to hide config output.

I should not reply to github messages late on Sundays I guess ;) and thanks to @sej7278 for catching up my mistake :)

@trombonehero I am closing it for now since we all agreed that this is already available.

PR #317, opened 2015-02-10
Updated Fedora spec file

suvayu:
Hi,

I updated the Fedora spec file as per latest Fedora packaging guidelines.  And made the necessary changes so `ccache` is used.  I have a test build here: https://copr.fedoraproject.org/coprs/fatka/arduino-mk/build/69001/.  Please feel free to give any feedback.

Cheers,

### DIFF
--- a/packaging/fedora/arduino-mk.spec
+++ b/packaging/fedora/arduino-mk.spec
@@ -1,13 +1,12 @@
 Name:			arduino-mk
 Version:		1.3.4
-Release:		1%{dist}
+Release:		2%{dist}
 Summary:		Program your Arduino from the command line
 Packager:		Simon John <git@the-jedi.co.uk>
 URL:            https://github.com/sudar/Arduino-Makefile
 Source:         %{name}-%{version}.tar.gz
 Group:			Development/Tools
 License:		LGPLv2+
-BuildRoot:		%{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 BuildArch:		noarch
 Requires:		arduino-core pyserial
 BuildRequires:	arduino-core
@@ -28,13 +27,11 @@ Arduino platform.
 mkdir -p %{buildroot}/%{_datadir}/arduino
 mkdir -p %{buildroot}/%{_bindir}
 mkdir -p %{buildroot}/%{_mandir}/man1
-mkdir -p %{buildroot}/%{_docdir}/%{name}/examples
-install -m 755 -d %{buildroot}/%{_docdir}/%{name}
-install -m 755 -d %{buildroot}/%{_docdir}/%{name}/examples
-for dir in `find examples -type d` ; do install -m 755 -d %{buildroot}/%{_docdir}/%{name}/$dir ; done
-for file in `find examples -type f ! -name .gitignore` ; do install -m 644 $file %{buildroot}/%{_docdir}/%{name}/$file ; done
+mkdir -p %{buildroot}/%{_pkgdocdir}/examples
+for dir in `find examples -type d` ; do install -m 755 -d %{buildroot}/%{_pkgdocdir}/$dir ; done
+for file in `find examples -type f ! -name .gitignore` ; do install -m 644 $file %{buildroot}/%{_pkgdocdir}/$file ; done
 install -m 644 *.mk arduino-mk-vars.md %{buildroot}/%{_datadir}/arduino
-install -m 644 licence.txt %{buildroot}/%{_docdir}/%{name}
+install -m 644 licence.txt %{buildroot}/%{_pkgdocdir}
 install -m 755 bin/ard-reset-arduino %{buildroot}/%{_bindir}/ard-reset-arduino
 install -m 644 ard-reset-arduino.1 %{buildroot}/%{_mandir}/man1
 
@@ -47,11 +44,13 @@ rm -rf %{buildroot}
 %{_mandir}/man1/ard-reset-arduino.1*
 %{_datadir}/arduino/*.mk
 %{_datadir}/arduino/arduino-mk-vars.md
-%doc %{_docdir}/%{name}/licence.txt
-%docdir %{_docdir}/%{name}/examples
-%{_docdir}/%{name}/examples
+%doc %{_pkgdocdir}/licence.txt
+%doc %{_pkgdocdir}/examples
 
 %changelog
+* Sat Jan 10 2015 Suvayu Ali <fatkasuvayu+linux@gmail.com>
+- Update according to latest Fedora packaging guidelines
+
 * Sat Apr 12 2014 Simon John <git@the-jedi.co.uk>
 - Put manpage back.
 * Fri Apr 04 2014 Simon John <git@the-jedi.co.uk>
--- a/packaging/fedora/arduino-mk.spec
+++ b/packaging/fedora/arduino-mk.spec
@@ -1,6 +1,6 @@
 Name:			arduino-mk
 Version:		1.3.4
-Release:		2%{dist}
+Release:		3%{dist}
 Summary:		Program your Arduino from the command line
 Packager:		Simon John <git@the-jedi.co.uk>
 URL:            https://github.com/sudar/Arduino-Makefile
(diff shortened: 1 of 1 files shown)


packaging/fedora/arduino-mk.spec
```diff
@@ -25,18 +24,21 @@ Arduino platform.
...
+mkdir -p %{buildroot}%{_pkgdocdir}/examples
+for dir in `find examples -type d` ; do install -m 755 -d %{buildroot}%{_pkgdocdir}/$dir ; done
+for file in `find examples -type f ! -name .gitignore` ; do install -m 644 $file %{buildroot}%{_pkgdocdir}/$file ; done
+install -m 644 *.mk arduino-mk-vars.md %{buildroot}%{_datadir}/arduino
+install -m 644 licence.txt %{buildroot}%{_pkgdocdir}
+install -m 755 bin/ard-reset-arduino %{buildroot}%{_bindir}/ard-reset-arduino
+install -m 644 ard-reset-arduino.1 %{buildroot}%{_mandir}/man1
+
+# Fedora uses ccache
+sed -i -e 's,\(CC \+= \)\($(AVR_TOOLS_PATH)\)\(/$(CC_NAME)\),\1/usr/lib64/ccache\3,' \
```
sej7278: should probably be ```%{_libdir}/ccache``` or ```%{_usr}/%{_lib}/ccache``` instead of ```/usr/lib64/ccache```

https://fedoraproject.org/wiki/Packaging:RPMMacros?rd=Packaging/RPMMacros
https://en.opensuse.org/openSUSE:Packaging_Conventions_RPM_Macros#.25_lib

also are you sure you should be removing the slashes everywhere?

```%{_pkgdocdir}``` only works for fedora 18 onwards

Comment (suvayu) [inline packaging/fedora/arduino-mk.spec]: Hi, you are correct both about the `ccache` lines, and `%{_pkgdocdir}`.  Since F17 (last version without `%{_pkgdocdir}`) was EOL'ed two years ago, it should be fine I think.

About the slashes, I'm pretty sure there should not be any.  But I'll double check nonetheless and report back.

Thanks for reviewing the changes. :)

**Comment by sej7278** [comment]
Did you get anywhere with this @suvayu ?

**Comment by suvayu** [comment]
@sej7278 Extremely sorry about the long delay.  Real life got a bit too hectic.

About the slashes, I could not find anything definitive, both seems to work.  So I reintroduced them to be on the safe side. Regarding using macros in the `ccache` paths, I updated it to use `%{_libdir}`.

I think this should cover everything.  I pushed the updated commits (overwriting my old ones), I see Github has picked that up, and updated the pull request!

**Comment by sej7278** [comment]
there's still a problem with the AVR_TOOLS_PATH detection, as we use a call to ```$(shell which $(CC_NAME))``` to find avr-gcc, and not all of the binaries are in the ccache location:
```
[simon@vbfedora20 blink]$ whereis avr-gcc
avr-gcc: /usr/bin/avr-gcc /usr/lib64/ccache/avr-gcc /usr/share/man/man1/avr-gcc.1.gz

[simon@vbfedora20 blink]$ whereis avr-ar
avr-ar: /usr/bin/avr-ar /usr/share/man/man1/avr-ar.1.gz

[simon@vbfedora20 blink]$ which avr-ar
/usr/bin/avr-ar

[simon@vbfedora20 blink]$ which avr-gcc
/usr/lib64/ccache/avr-gcc
```

so we end up with the binaries that aren't using ccache, not being found:
```
make: /usr/lib64/bin/avr-ar: Command not found
make: *** [build-mega2560/libcore.a] Error 127
[simon@vbfedora20 blink]$ which avr-ar 
/usr/bin/avr-ar
```

and avr-size is not found:
```
Arduino.mk Configuration:
- [AUTODETECTED]       CURRENT_OS = LINUX 
- [AUTODETECTED]       ARDUINO_DIR = /usr/share/arduino 
- [COMPUTED]           ARDMK_DIR = /usr/share/arduino (relative to Common.mk)
- [AUTODETECTED]       ARDUINO_VERSION = 105 
- [DEFAULT]            ARCHITECTURE =  
- [DEFAULT]            VENDOR = arduino 
- [DEFAULT]            ARDUINO_SKETCHBOOK = /home/simon/sketchbook 
- [AUTODETECTED]       AVR_TOOLS_DIR = /usr/lib64 (found in $PATH)
- [COMPUTED]           ARDUINO_LIB_PATH = /usr/share/arduino/libraries (from ARDUINO_DIR)
- [COMPUTED]           ARDUINO_VAR_PATH = /usr/share/arduino/hardware/arduino//variants (from ARDUINO_DIR)
- [COMPUTED]           BOARDS_TXT = /usr/share/arduino/hardware/arduino//boards.txt (from ARDUINO_DIR)
- [DEFAULT]            USER_LIB_PATH = /home/simon/sketchbook/libraries (in user sketchbook)
- [DEFAULT]            PRE_BUILD_HOOK = pre-build-hook.sh 
- [USER]               BOARD_TAG = mega2560 
- [COMPUTED]           CORE = arduino (from build.core)
- [COMPUTED]           OBJDIR = build-mega2560 (from BOARD_TAG)
- [COMPUTED]           ARDUINO_CORE_PATH = /usr/share/arduino/hardware/arduino//cores/arduino (from ARDUINO_DIR, BOARD_TAG and boards.txt)
- [ASSUMED]            MONITOR_BAUDRATE = 9600 
- [DEFAULT]            OPTIMIZATION_LEVEL = s 
- [DEFAULT]            MCU_FLAG_NAME = mmcu 
- [DEFAULT]            CFLAGS_STD =  
- [DEFAULT]            CXXFLAGS_STD =  
- [COMPUTED]           DEVICE_PATH = /dev/ttyACM0 (from MONITOR_PORT)
- [DEFAULT]            FORCE_MONITOR_PORT =  
make: /usr/lib64/bin/avr-size: Command not found
- [AUTODETECTED]       Size utility: Basic (not AVR-aware)
- [COMPUTED]           BOOTLOADER_PARENT = /usr/share/arduino/hardware/arduino//bootloaders (from ARDUINO_DIR)
- [COMPUTED]           ARDMK_VERSION = 1.3.4 
- [COMPUTED]           CC_VERSION = 4.8.2 (avr-gcc)
```

it is a bizarre way of using ccache that fedora have gone with!

**Comment by sej7278** [comment]
changing the binary we're looking for in which() to avr-as, seems to fix the issue - it finds the tools in /usr/bin/ but still uses /usr/lib64/ccache/avr-g++ and avr-gcc.

the problem is that avr-as is provided by the package avr-binutils on fedora and binutils-avr on debian, and they're not dependencies of arduino or arduino-core, so not automatically installed with the ide or avr-gcc etc; so we can't rely on the command being present.

also fedora adds copies all of the/usr/bin/avr* binaries into /usr/avr/bin/ too, so that's 3 places to look now!

so it seems we either stick with the current workaround of specifying the paths to all the binaries with ```OVERRIDE_EXECUTABLES = 1``` or edit the specfile to patch Arduino.mk just for fedora, which i'm not keen on.

```diff
diff --git a/Arduino.mk b/Arduino.mk
index a5b9bba..d5f39da 100644
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -429,7 +429,7 @@ ifndef AVR_TOOLS_DIR
 
     else
 
-        SYSTEMPATH_AVR_TOOLS_DIR := $(call dir_if_exists,$(abspath $(dir $(shell which $(CC_NAME)))/..))
+        SYSTEMPATH_AVR_TOOLS_DIR := $(call dir_if_exists,$(abspath $(dir $(shell which $(AR_NAME)))/..))
         ifdef SYSTEMPATH_AVR_TOOLS_DIR
             AVR_TOOLS_DIR = $(SYSTEMPATH_AVR_TOOLS_DIR)
             $(call show_config_variable,AVR_TOOLS_DIR,[AUTODETECTED],(found in $$PATH))
```

suvayu:
Hi, I think avr-binutils is required by arduino on Fedora, but indirectly.  Take a look at the following:

    $ rpm -q --whatrequires avr-binutils
    avr-gcc-4.9.2-1.fc21.x86_64
    $ rpm -q --whatrequires avr-gcc
    arduino-core-1.0.5-8.fc21.noarch
    arduino-doc-1.0.5-8.fc21.noarch
    avr-gcc-c++-4.9.2-1.fc21.x86_64
    $ rpm -q --requires arduino | grep arduino
    arduino-core = 1:1.0.5-8.fc21
    arduino-doc = 1:1. rpm -q --whatrequires avr-binutils
    avr-gcc-4.9.2-1.fc21.x86_64

Does that make it better?  I also looked at the issue of copies of binaries being in `/usr/avr/bin` and `/usr/bin/avr-*`, it turns out Fedora doesn't do that explicitly.  The Makefile from up stream does that.  So I think you can simply rely on the output of `which`, or did I misunderstand something?

sej7278:
have you tried to build something with the makefile - it doesn't work as ```which gcc``` points everything at /usr/lib64 which means half of the avr tools are not found as they're in /usr/bin

suvayu:
Hi, sorry I can't seem to find the time to do all the testing this deserves.  I'm closing the pull request now.  When I have more time, I'll open a new pull request after diligent testing.

PR #322, opened 2015-02-26
Add "linuxgpio" to the list of ISP's that don't have a port

onandoffables:
Hi,

I've been using avrdude-6.1 (compiled from source with linuxgpio enabled) to program on the Raspberry Pi. When using 'make ispload', avrdude complains about the port. Just like these:

gpio pull request: https://github.com/sudar/Arduino-Makefile/pull/166
avrispmkii pull request: https://github.com/sudar/Arduino-Makefile/pull/282

Adding "linuxgpio" to the list of ISP's that don't have a port fixes the issue. Linuxgpio is now a built-in option in avrdude.

I would really appreciate it if you added it to your awesome Makefile. Thanks!

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1342,7 +1342,7 @@ endif
 AVRDUDE_ISP_OPTS = -c $(ISP_PROG) -b $(AVRDUDE_ISP_BAUDRATE)
 
 ifndef $(ISP_PORT)
-    ifneq ($(strip $(ISP_PROG)),$(filter $(ISP_PROG), usbasp usbtiny gpio avrispmkii))
+    ifneq ($(strip $(ISP_PROG)),$(filter $(ISP_PROG), usbasp usbtiny gpio linuxgpio avrispmkii))
         AVRDUDE_ISP_OPTS += -P $(call get_isp_port)
     endif
 else


sej7278:
has gpio been renamed to linuxgpio now, or is it something different?

onandoffables:
'linuxgpio' is different from 'gpio', even though they do the same thing with the same pins on the same device.

'linuxgpio' comes with upstream avrdude-6.1 (at least). It's enabled when compiled with './configure --enable-linuxgpio=yes'. It works out of the box on a Raspberry Pi (after editing avrdude.conf to assign pins), but also on non-Raspberry Pi boards. And there are plenty of those!

'gpio' dates from when 'linuxgpio' wasn't in upstream avrdude yet. Our friend Gordon from the Raspberry Pi forums took avrdude-5.10 (the latest at the time) and patched it to work with the Raspberry Pi GPIO's. We were all happy and life was good. When https://github.com/sudar/Arduino-Makefile/pull/166 was added to this repository, life was even better.

Pretty much everyone is using that avrdude-5.10 + avrdude.conf on the Raspberry Pi, but it may be time to move on now that's it's a standard option in upstream avrdude.

Also, I haven't seen the avrdude-5.10 gpio source code / patch anywhere, which leaves us with a binary blob. That's also problematic for some people. For example, those that run avrdude and Arduino-Makefile on their Odroids, Beagles, Bananas, Hummingboards and what have you.

sej7278:
looks good to me, and thanks for the explanation :+1:

sudar:
@onandoffables Thanks for the pull request. Just merged it.

@sej7278 Thanks for validating it.

onandoffables:
Thank you both!

For anyone interested in using avrdude-6.1 with linuxgpio on their linux board; I have a very simple script that downloads, unpacks, patches (avrdude.conf), compiles and installs it here: https://github.com/onandoffables/avrdude-linuxgpio

It's all simple and self-explanatory, it just saves typing. The instructions assume a Debian based Linux distro (such as Raspbian), but that's easily adapted.

sej7278:
interesting, have you thought of making a (unofficial) debian package of that, makes it a it easier to uninstall or upgrade if raspbian ever moves past wheezy, and if you've got multiple devices you don't need to download/compile every time then.

i did something similar with dfu-util as i needed a more modern branch for my spark core and wanted to run it on multiple dekstops, pi's and a bbb

onandoffables:
I'll look into that, thanks!

It takes about 3 minutes to compile on a new Raspberry Pi model 2, and I'm afraid there'll always be *some* manual/scripted labour involved in assigning GPIO pins in avrdude.conf. I also have a script (part of the bigger plan) that gives the option of simply copying a pre-compiled avrdude binary and ready-to-go config.

But I agree that having a debian package (preferably in the repo) would be best, yes. Then you can also skip installing of prerequisites to be able to compile it in the first place, and save more time.

Thanks for the tip!

sej7278:
actually, i'm 99% done with a debian package, just waiting for the pi to build it which is waaaaay more than 3mins on B (1)

onandoffables:
Good stuff, now we're getting somewhere :)

I'll test both packages on a clean SD card, but need some time for that. Hopefully later tonight or tomorrow.

Thanks again, really appreciated!

onandoffables:
This is probably not the best place to discuss this (sorry @sudar), but both your packages work fine @sej7278.

Regarding the problem with avrdude.conf: The debian avrdude package maintainers have no way of knowing what pins to choose from what numbering scheme. Just leaving the pins question-marked and commented in the config and have avrdude generate an error is probably the only sensible thing to do. I can understand having pins 8,9,10,11 enabled/set in Raspbian though.

There's a make_deb.sh script in the repo (https://github.com/onandoffables/avrdude-linuxgpio) now that creates an avrdude_6.1-2_armhf.deb. Probably crude, but the result is very similar to yours and the one from debian you posted. Except both yours and the one from debian depend on libelf1 (which the current 5.11 one from the Raspbian repo doesn't), so I compiled without it and removed that dependency.

Now avrdude can be installed with just libftdi1 (which is installed by the Arduino package anyway) on a clean SD card and it all works well.

Your 1.3.4-2 arduino-mk package works wonderfully. Thanks!

PR #329, opened 2015-03-25
Added search of PLATFORM_LIB_NAMES to ARDUINO_LIBS

sej7278:
....so that Wire.h is found on 1.6.1 and is output separately to SYS_LIB_NAMES

Added "sort" to the USER/SYS/PLATFORM _LIBS so we don't get duplicates


sudar:
I was about to update the `History.md` file and then found that you remembered :)

Travis CI also passed. So merging it.

PR #331, opened 2015-03-26
Added BOARD_SUB to OBJDIR

sej7278:
So that when changing sub boards we don't accidentally overwrite existing builds.

E.g. mega1280 and mega2560 previously *both* created a "build-mega" directory, now they'll create build-mega-mega2560 and build-mega-mega1280

Only applies to 1.5+ of course, and only when BOARD_SUB is used. 1.0 will still create build-mega2560 and build-mega1280 directories (just BOARD_TAG)

Also, OBJDIR is still completely user-overridable, we're not forcing the build-TAG-SUB format on them


sudar:
Thanks. Just merged it.

PR #333, opened 2015-03-27
Add = to PARSE_BOARD regex to make it less greedy.

sej7278:
For instance in the Sparkfun 1.6 core, we have:

```
  promicro16.build.vid.0=0x1B4F
  promicro16.build.vid.1=0x1B4F
  promicro16.build.vid=0x1B4F
```

So we end up matching all 3 instead of just the last one.

Adding the = means we're looking for ```promicro16.build.vid=``` so not catching the .0 or .1 version.

Probably needs some good discussion as its quite an important function, but I can't really see it breaking things as we already pipe to ```cut -d =```, the only possibility is if there's whitespace around the = i guess, but not sure that's a valid boards.txt entry then anyway.

See details about multiple usb vid/pid's here: https://github.com/arduino/Arduino/wiki/Arduino-Hardware-Cores-migration-guide-from-1.0-to-1.6#step-6-add-usb-identifiers-for-port-list-menu

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -552,7 +552,7 @@ endif
 
 ifndef PARSE_BOARD
     # result = $(call READ_BOARD_TXT, 'boardname', 'parameter')
-    PARSE_BOARD = $(shell grep -v '^\#' $(BOARDS_TXT) | grep $(1).$(2) | cut -d = -f 2 )
+    PARSE_BOARD = $(shell grep -v '^\#' $(BOARDS_TXT) | grep $(1).$(2)= | cut -d = -f 2 )
 endif
 
 # If NO_CORE is set, then we don't have to parse boards.txt file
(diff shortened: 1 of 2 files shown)


sej7278:
Also a little tweak to auto-detect the Arduino 1.5+ preferences.txt on Linux at least.

sudar:
My talk is in a couple of hours. I will merge it after that.

sej7278:
How did the talk go?

sudar:
That talk was good. Here are the slides. http://hardwarefun.com/news/programming-arduino-using-makefile

Most of them were beginners so they were not able to appreciate it, but there were a few people who got it and told me that they will try it out and let me know if they have any feedback.

sej7278:
cool, i'll have a look at the slides in a mo. so are we going to do an arduino day release?

sudar:
> so are we going to do an arduino day release?

If you feel that we don't need to add anything more, then we can make an "Arduino Day" Release.

PR #334, opened 2015-03-28
Fixed typo - bracket should be backslash to continue the call to firstword

sej7278:


sudar:
Oops!

Thanks for fixing it.

PR #337, opened 2015-03-29
Replaced sub-make with function calls

michaelbaisch:
By not using sub-make but function calls one is able to use this as a global makefile.

Let's say you want to use this makefile in a script or app and don't want to have a makefile in every project folder.
This could be done like this:

```
# Set environment variables
$ cd ~/my-arduino-project
$ make -f ~/Arduino-makefile/Arduino.mk
```

But using `upload` isn't working because it's using sub-make:

```
$ make -f ~/Arduino-makefile/Arduino.mk upload
make[1]: *** No rule to make target `reset'.  Stop.
make: *** [upload] Error 2
```

Is the use of sub-make appropriate here? Because no separate makefile is used.

https://www.gnu.org/software/make/manual/html_node/Recursion.html

> This technique is useful when you want separate makefiles for various subsystems that compose a larger system. For example, suppose you have a subdirectory ‘subdir’ which has its own makefile, and you would like the containing directory’s makefile to run make on the subdirectory.
> 
> subsystem:
> &nbsp;&nbsp;&nbsp;&nbsp;cd subdir && $(MAKE)

The problem is that the -f flag isn't passed down, which makes sense if you use sub-make for separate makefiles.

https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html

> The options ‘-C’, ‘-f’, ‘-o’, and ‘-W’ are not put into MAKEFLAGS; these options are not passed down.

My suggestion is to use functions instead of sub-make.
I'm aware that this is a specific problem and maybe there is a cleaner solution.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1406,39 +1406,47 @@ $(TARGET_ELF): 	$(LOCAL_OBJS) $(CORE_LIB) $(OTHER_OBJS)
 $(CORE_LIB):	$(CORE_OBJS) $(LIB_OBJS) $(PLATFORM_LIB_OBJS) $(USER_LIB_OBJS)
 		$(AR) rcs $@ $(CORE_OBJS) $(LIB_OBJS) $(PLATFORM_LIB_OBJS) $(USER_LIB_OBJS)
 
-error_on_caterina:
-		$(ERROR_ON_CATERINA)
+error_on_caterina: $(call error_on_caterina)
 
+error_on_caterina =	$(ERROR_ON_CATERINA)
 
 # Use submake so we can guarantee the reset happens
 # before the upload, even with make -j
 upload:		$(TARGET_HEX) verify_size
-		$(MAKE) reset
-		$(MAKE) do_upload
+		$(call reset)
+		$(call do_upload)
 
 raw_upload:	$(TARGET_HEX) verify_size
-		$(MAKE) error_on_caterina
-		$(MAKE) do_upload
+		$(call error_on_caterina)
+		$(call do_upload)
 
 do_upload:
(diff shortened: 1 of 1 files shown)


sudar:
@michaelbaisch Thanks for the suggestion and the pull request.

While on the whole I may agree with you that we should use functions instead of sub-make, it may not be possible for us to change it in the current project because of the following reasons.

- There are lot of customisation options that needs to be set and it may not be practical to specify all of them in the command line each time.
- This project has been for a while and there are lot of users who are already using it. Changing it to use functions might break things for them.

sej7278:
also there is a comment that submake has been used to ensure that reset happens before upload when using multiple threads to compile

sudar:
@michaelbaisch As @sej7278 mentioned "making sure reset happens before upload" is the third reason why we are using sub-make.

Because of the above mentioned reasons, we will not be able to merge this. But if you have a solution which address these issues then we would be happy to look into it again.

michaelbaisch:
@sudar As far as I know by using functions like I did in the commit everything should work like before. And the ability to use this makefile as a global makefile (with specifying all environment variables in the command line) is with this also possible.

But as I looked into it I understand @sej7278 concern about multiple threads. 
My makefile knowledge is rather rudimentary but shouldn't it be possible to use "order-only prerequisites" to ensure the right order? And maybe the additional functions aren't even necessary because the targets could be called directly?
http://www.gnu.org/software/make/manual/make.html#index-order_002donly-prerequisites

sudar:
@michaelbaisch 

Are you still interested in getting this feature merged?

If yes, then please refresh your PR and we can start doing some tests.

I am closing this for now.  But feel free to re-open when you refresh the PR.

PR #339, opened 2015-04-04
Better autodetection of ARDUINO_SKETCHBOOK and ARDUINO_DIR on OSX

sej7278:
Added 1.6 path for ARDUINO_DIR and also a better OSX default for ARDUINO_SKETCHBOOK than just the Linux $(HOME)/sketchbook

So we can now just have a Makefile consisting of:

```
BOARD_TAG = uno 
MONITOR_PORT = /dev/ttyACM0

include $(HOME)/arduino-mk/Arduino-Makefile/Arduino.mk
```

And you don't need to run the IDE to create preferences.txt to find the sketchbook location (if you use the default).


sudar:
Thanks @sej7278 :)

PR #340, opened 2015-04-04
Fix reset for leonardo

calvinli:
I found that the previous reset code did not work for Leonardo or Micro, nor did any reset code anywhere on the internet.

An examination of the Arduino IDE source code (and the JNI code behind its serial implementation) shows that it holds RTS high and DTR low when doing the 1200 baud open/close. And it turns out that's exactly what's needed!

### DIFF
--- a/bin/ard-reset-arduino
+++ b/bin/ard-reset-arduino
@@ -17,10 +17,10 @@ if args.caterina:
     if args.verbose: print('Forcing reset using 1200bps open/close on port %s' % args.port[0])
     ser = serial.Serial(args.port[0], 57600)
     ser.close()
-    ser.open()
-    ser.close()
     ser.setBaudrate(1200)
     ser.open()
+    ser.setRTS(True)  # RTS line needs to be held high and DTR low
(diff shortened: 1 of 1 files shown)


calvinli:
A correction: it is the JSSC source code, not the Arduino source, that contains the RTS/DTR weirdness: https://github.com/scream3r/java-simple-serial-connector/

sej7278:
@calvinli Brilliant!

This definitely works for me when the old method won't. Tried with 1.0.5 and 1.6.3 with Leonardo and Sparkfun Pro Micro.

I do recall reading in the 1.6 releasenotes that they'd ditched the old reset method in favour of doing it with JSSC (as the 1.6.3 IDE resets fine too now whereas 1.0.5 didn't always) as Debian has only just started packaging it.

@sudar I wouldn't be so bold as to merge an external PR myself, but this looks good to me (needs a note in HISTORY.md)

sudar:
@calvinli 

Thanks for the pull request. We have been having this issue with reset for a really long time.

@sej7278 

Thanks for testing the patch. Just merged it.

Issue #251, opened 2014-09-06
Link to cpp files in subfolders


roobre:
It looks like `cpp` files in subfolders are ignored.

Would be possible to get these compiled and linked too? Subfoldering really helps when having lots of files :\

Regards.

ladislas:
@roobre any news on the issue?

sudar:
@roobre I am closing this ticket now because of inactivity.

Feel free to re-open if needed.

martin-mh:
Hi everyone !
Same problem here.
I'm working on a medium sized arduino project. Here the explaination : the project use a LCD. To write on it, i've cut different states in something called "Menu". Each Menu is a class who define the behavior of each menus. Since a Menu is not the "core logic" of the app, i putted them into separated folder.
Headers files are compiled because i call them but it tell me an compilation error due to reference error on functions who are defined in .cpp.
And more I can't include cpp files from the header because i'm using forwarded class declarations.

Take a look here : https://github.com/martin-mh/Projet-SI-2015-2016/tree/a485d052c70d383b2369b381afd3dc0267d2156e

So i got to move everything in the root folder ! Here : https://github.com/martin-mh/Projet-SI-2015-2016

Do you have any work around or any fix for that ? Because having every files in the root folder is ugly, in my mind.
Thanks.

sej7278:
given that arduino-builder has just implemented recursively search subdirectories and totally broken itself (as it tries to link to the existing precompiled temp files!) i don't think its a good idea

martin-mh:
ah darn ! That's not a big deal in anyway, don't worry :p

Overdrivr:
Any updates on that ? I would like as well to put source/headers files inside subfolders to organize code inside the library.

Pharap:
This is still an issue.

Having classes (note: `class`es, not libraries) split into `.h` and `.cpp` files might not be 'idiomatic Arduino', but it is completely expected idiomatic C++.

By not handling such cases, the IDE is crippling the experienced programmers who are using well-established good coding practices.

roobre:
For anyone still wondering how to do this, I just uglily script it inside my makefile:

```makefile
# This variable is local, name it however you want
SOURCE_FOLDERS = src src/Player src/Weapon

# These two are used by Arduino.mk
LOCAL_C_SRCS = $(foreach dir,$(SOURCE_FOLDERS),$(wildcard $(dir)/*.c))
[...]

Issue #249, opened 2014-09-01
Local libraries in sketch directory


ladislas:
Hi guys!

I've made some test. It seems that if I put my `.h` and  `.cpp` files in my sketch folder, it compiles, but if I put them inside a `lib` directory inside the sketch folder, it does not work...

So this works:

``` bash
.
├── Blink.ino
├── Makefile
├── MyFirstLib.cpp
└── MyFirstLib.h
```

But this does not:

``` bash
.
├── Blink.ino
├── Makefile
└── lib
    ├── MyFirstLib.cpp
    └── MyFirstLib.h
```

It this normal? :)

PR #341, opened 2015-04-05
Add automatic lib detection

ladislas:
Hey there!

It's been a while, but I'm back with my *automatic library detection PR* ;)

I've tested it on Bare Arduino Project and Moti and it works great.

Would love to have some feedbacks :)

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -784,7 +784,7 @@ ifeq ($(strip $(CHK_SOURCES)),)
                 $(call show_config_info,No .pde or .ino files found. If you are compiling .c or .cpp files then you need to explicitly include Arduino header files)
             else
                 #TODO: Support more than one file. https://github.com/sudar/Arduino-Makefile/issues/49
-                $(error Need exactly one .pde or .ino file. This makefile doesn't support multiple .ino/.pde files yet)
+                $(error Need exactly one .pde or .ino file. This makefile doesn\'t support multiple .ino/.pde files yet)
             endif
         endif
 
@@ -812,6 +812,19 @@ else
     $(call show_config_info,NO_CORE set so core library will not be built,[MANUAL])
 endif
 
+
+########################################################################
+# Automatically find the libraries needed to compile the sketch
+
+ifndef SKETCH_LIBS
+	SKETCH_LIBS := $(shell $(ARDMK_DIR)/bin/lib-detection $(USER_LIB_PATH) | \
+                sed -ne 's/SKETCH_LIBS \(.*\) /\1/p')
+endif
+
+ifndef SKETCH_LIBS_DEPS
+	SKETCH_LIBS_DEPS := $(shell $(ARDMK_DIR)/bin/lib-detection $(USER_LIB_PATH) | \
+                sed -ne 's/SKETCH_LIBS_DEPS \(.*\) /\1/p')
+endif
 
 ########################################################################
 # Determine ARDUINO_LIBS automatically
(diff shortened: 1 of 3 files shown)


sej7278:
the main thing that bugs me is shelling out to another python script :-(

what does this do that the current detection doesn't - libs that call other libs or something? as system, platform and user lib autodetection already works quite well i think.

can you provide some build output so we can get an idea?

ladislas:
> libs that call other libs or something?

Yes exactly. The idea is to be able to only include the main lib in your sketch and let the makefile figure out which other libs it needs to link and compile.

This is something the IDE is going to support real soon.

Here is the output of the build process: https://gist.github.com/ladislas/b5d245e9f8f4ffa6a26a

sej7278:
i'm surprised it doesn't remove a shedload of code from Arduino.mk then, for instance all of this lot becomes redundant doesn't it?

https://github.com/sudar/Arduino-Makefile/blob/master/Arduino.mk#L825
https://github.com/sudar/Arduino-Makefile/blob/master/Arduino.mk#L916
https://github.com/sudar/Arduino-Makefile/blob/master/Arduino.mk#L1154

does it still output the object files to different directories like userlibs, platformlibs, libs, core etc?

another two things i can think of - we'd need to search for the binary in the path, not just ARDMK_DIR/bin, like we do here (as debian etc. would want it in /usr/bin/):

https://github.com/sudar/Arduino-Makefile/blob/master/Arduino.mk#L737

and also it would need a manpage, which also means we'd have to update the fedora specfile (and the debian package).

sej7278:
this works for me (ignoring tabs->spaces):

```diff
diff -w lib-detection ~/lib-detection 
6a7
> 
11c12
< includeRegex = re.compile('^\#include\s(<|")(.*)\.h("|>)', re.DOTALL|re.M)
---
> includeRegex = re.compile("(?<=^\#include\s\")(.*)(?=\.h\")", re.DOTALL|re.M)
43c44
<             group = match.group(2)
---
> 			group = match.group(1)
57c58
<                 group = match.group(2)
---
> 				group = match.group(1)
72c73
<                     group = match.group(2)
---
> 					group = match.group(1)
```

sej7278:
Here's my "fixes": https://github.com/sej7278/Arduino-Makefile/tree/auto-lib

ladislas:
@sej7278 

thanks for all your feedbacks! :) I'm currently working on it and to answer your questions:

> does it still output the object files to different directories like userlibs, platformlibs, libs, core etc?

yes.

> we'd need to search for the binary in the path, not just ARDMK_DIR/bin

Ok, I'll do that.

> which probably means giving it a more specific name than "lib-detection", maybe "ard-lib-detection"?

Changing that.

> and replace tabs with spaces in the python script.

sure.

> and the regex only matches quotes not brackets, so this doesn't work: #include <mylib.h> -- hang on, does this only work on user libs?

yes, the idea was to only work on user libs and not on the Arduino core and platform libs. **Do you think it could be useful as well?** 

> it seems to be replacing USER_LIBS and ignoring ARDUINO_LIBS if manually set.

I'll look into that as well as into the *useless* code.

> Can we close PR #291 ?

Already done.

> Does this fix any of these issues?

Yes for #93 and #251. Haven't tried #249 yet.

> Either way, we should probably tag a new release before merging this sort of big change in.

Agreed

sej7278:
@ladislas - if you look at my branch linked above, i've already fixed most of the items other than removing "legacy" code and writing a manpage. see what you think.

as far as issue #251 goes, its only fixed if USER_LIB_PATH is set to your sketch directory, otherwise it doesn't find libraries storied within the sketch directory.

i don't think its worth supporting platform/system libs, it'll just slow things down more and they shouldn't change often anyway. might be worth documenting though.

Issue #346, opened 2015-04-12
BOARD_TAG "atmegang" with BOARD_SUB isn't building


michaelbaisch:
In newer releases `atmegang` has two `BOARD_SUB` but the right `MCU` is never found by the makefile because the boards.txt has this line: `atmegang.build.mcu=atmegang`.
Can anybody image why this line is in the boards.txt? 
Obviously it's not even a valid argument for `-mmcu`.

The makefile first looks if `$(BOARD_TAG),build.mcu` exists and afterwards if `$(BOARD_TAG),menu.cpu.$(BOARD_SUB).build.mcu` exists.
 It works if these are switched:
```
# processor stuff
ifndef MCU
    MCU := $(call PARSE_BOARD,$(BOARD_TAG),menu.cpu.$(BOARD_SUB).build.mcu)
    ifndef MCU
        MCU := $(call PARSE_BOARD,$(BOARD_TAG),build.mcu)
    endif
endif
```

Question is if it would, in general, make more sense to first check if something with `BOARD_SUB` exists?

sej7278:
it does seem pretty odd that they have build.mcu as atmegang which isn't a valid mcu, and then the cpu submenu is being used as the mcu.

the current method works for nano's and mega's, your patch breaks nano's as the mcu in the submenu is invalid, and there is no build.mcu defined. so we need something that can handle these examples:

```
nano.menu.cpu.atmega328.build.mcu=atmega328p
mega.menu.cpu.atmega2560.build.mcu=atmega2560
atmegang.menu.cpu.atmega168.build.mcu=atmega168
```

PR #347, opened 2015-04-12
Looks for submenu items first when parsing BOARDS_TXT

sej7278:
As suggested by @michaelbaisch in issue #346

@sudar might want to update HISTORY.md with an "in development" version number.


sej7278:
didn't really need to change the caterina detection, but it matches the other methods, so looks cleaner i guess.

michaelbaisch:
Thanks for the patch.

PR #354, opened 2015-05-21
Added some more advanced tests to Travis CI

sej7278:
Better than just "make clean && make". 

Tested on Arduino, Teensy and chipKIT, they all provide the functions called.

Could probably do with adding IDE 1.6.4 tests next (alongside 1.0.6 not instead of).

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -8,6 +8,7 @@ I tried to give credit whenever possible. If I have missed anyone, kindly add it
 
 - Tweak: Looks for submenu items first when parsing BOARDS_TXT (Issue #347) (https://github.com/sej7278)
 - Tweak: Various spelling/grammar/typo fixes (https://github.com/dcousens)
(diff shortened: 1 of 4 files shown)


sudar:
@sej7278 Thanks.

I am just waiting for Travis CI to complete before merging this.

sej7278:
well the AVR C example failed as it skips the assembly stage (no core to link) but Travis_CI still passed it, so not sure what use CI is!

```
make: *** No rule to make target `build-atmega16/BlinkInAVRC.s', needed by `generate_assembly'.  Stop.
```

sudar:
Hmm.. That's strange.

Do you me hold of merging this PR, till we figure it out?

sej7278:
trying again, but i've got to do a couple more commits to check it does break when it should, then i'll clean it up.

sej7278:
ok i checked it does break when expected to now. new PR is at #355 should be ok to merge

PR #357, opened 2015-05-31
Add support for attiny84a

hhgarnes:
Attiny84a requires two things:
-Support for rewriting the avrdude mcu to attiny84 even if the gcc mmcu parameter stays as attiny84a
-Removal of the avrdude -D flag. Without this, no uploaded sketch would ever take on the attiny84a (determined by experimentation and comparison of parameters with the arduino ide)

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1282,8 +1282,8 @@ ifndef AVRDUDE_OPTS
     AVRDUDE_OPTS = -q -V
 endif
 
-# Avrdude uses mcu attiny84 to upload to the attiny84a, but we want to be able to set
-# the gcc mmcu flag to attiny84a.
+# Decouple the mcu between the compiler options (-mmcu) and the avrdude options (-p).
+# This is needed to be able to compile for attiny84a but specify the upload mcu as attiny84.
 ifeq ($(MCU), attiny84a)
   AVRDUDE_MCU = attiny84
 else
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1290,21 +1290,14 @@ ifndef AVRDUDE_MCU
   AVRDUDE_MCU = $(MCU)
 endif
 
-# -D - Disable auto erase for flash memory
-# -D is needed for Mega boards. (See https://github.com/sudar/Arduino-Makefile/issues/114#issuecomment-25011005)
-# -D must NOT be present for attiny84a (unknown for attiny84)
-ifeq ($(AVRDUDE_MCU), attiny84)
-    AVRDUDE_ERASE_FLASH_OPT =
-else
-    AVRDUDE_ERASE_FLASH_OPT = -D
-endif
-
-AVRDUDE_COM_OPTS = $(AVRDUDE_OPTS) $(AVRDUDE_ERASE_FLASH_OPT) -p $(AVRDUDE_MCU)
+AVRDUDE_COM_OPTS = $(AVRDUDE_OPTS) -p $(AVRDUDE_MCU)
 ifdef AVRDUDE_CONF
     AVRDUDE_COM_OPTS += -C $(AVRDUDE_CONF)
 endif
 
-AVRDUDE_ARD_OPTS = -c $(AVRDUDE_ARD_PROGRAMMER) -b $(AVRDUDE_ARD_BAUDRATE) -P
+# -D - Disable auto erase for flash memory
+# -D is needed for Mega boards. (See https://github.com/sudar/Arduino-Makefile/issues/114#issuecomment-25011005)
+AVRDUDE_ARD_OPTS = -D -c $(AVRDUDE_ARD_PROGRAMMER) -b $(AVRDUDE_ARD_BAUDRATE) -P
 ifeq ($(CURRENT_OS), WINDOWS)
     # get_monitor_port checks to see if the monitor port exists, assuming it is
     # a file. In Windows, avrdude needs the port in the format 'com1' which is
@@ -1368,6 +1361,7 @@ ifndef AVRDUDE_ISP_FUSES_POST
     endif
 endif
 
+# -D may cause issues with some boards like attiny84a.
 AVRDUDE_ISP_OPTS = -c $(ISP_PROG) -b $(AVRDUDE_ISP_BAUDRATE)
 
 ifndef $(ISP_PORT)
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1282,14 +1282,26 @@ ifndef AVRDUDE_OPTS
     AVRDUDE_OPTS = -q -V
 endif
 
-AVRDUDE_COM_OPTS = $(AVRDUDE_OPTS) -p $(MCU)
+# We avrdude uses attiny84 to upload to the 84a, but we want to be able to set the gcc mmcu flag to attiny84a.
+ifeq ($(MCU), attiny84a)
+  AVRDUDE_MCU = attiny84
+else
+  AVRDUDE_MCU = $(MCU)
+endif
+
+# -D - Disable auto erase for flash memory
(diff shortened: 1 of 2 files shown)


sej7278:
i'm no fan of the -D option, i'd actually rather the exception be to only pass it for mega2560's which seems to be the only chip/board that requires it.

but what core are you using for attiny84a, and what's the mcu issue?

hhgarnes:
I'm using the Arduino-Tiny core. This patch came to be because I was trying everything possible to get my code to run on the attiny84a, and one of the initial things I saw was that it was compiling like this:
avr-g++ -MMD -c -mmcu=attiny84 instead of like this avr-g++ -MMD -c -mmcu=attiny84a, so I manually added some boards in the core that specified attiny84a as the mmc. However, this broke avrdude which didn't recognize attiny84 as a valid mcu. I updated the comment about that.

I'd be happy to change this to include -D for the mega only, but I can't really verify how to specify it. If you can confirm that I can change the ifneq to ifeq ($(AVRDUDE_MCU), mega2560), I'd be happy to do that.

sej7278:
i see, so avr-g++ has a ```attiny84a``` mcu, but avrdude only has ```attiny84```

i quite like the idea of splitting the compiler (etc.) mcu from avrdude mcu. i think i'd rather reduce your patch to:

```
indef AVRDUDE_MCU
  AVRDUDE_MCU = $(MCU)
endif

AVRDUDE_COM_OPTS = $(AVRDUDE_OPTS) -p $(AVRDUDE_MCU)
```

and then you can set the following in your local Makefile to set the mcu and remove the -D flag:

```
BOARD_TAG      = attiny84a
AVRDUDE_MCU    = attiny84
AVRDUDE_OPTS   = -v
```

we can't really remove the -D option as a default in Arduino.mk as we try to use the same flags as the [ide](https://github.com/arduino/Arduino/blob/57dcf07244629b4492ad49fb7c1188694a2907ef/hardware/arduino/avr/platform.txt#L94) and it does make sense to disable auto-erase to prolong the life of the flash (apart from the mega2560 issue).

hhgarnes:
I updated the mcu decoupling to your liking.

Your -D comment is a bit confusing. The IDE doesn't send the -D parameter, at least not for attiny84, only Arduino.mk does that. So the mega2560 would be the only board that has an issue with _removing_ -D (and thereby it's the only board that would require deviating from what the IDE does). It should at least be made obvious somewhere that -D needs to go for the attiny84a, or more people would be spending a lot of time debugging this. I updated the code to be more clear on this.

And yes, I had done so much hackery before I found the -D option that I had forgotten my patch to the tiny library, sorry about that. Here, for reference:

--- tiny_ardionoorig/avr/cores/tiny/core_build_options.h	2015-05-30 21:33:26.483497944 +0200
+++ tiny/avr/cores/tiny/core_build_options.h	2015-05-28 23:32:52.789650574 +0200
@@ -68,7 +68,7 @@
   Build options for the ATtiny84 processor
 =============================================================================*/
 
-#if defined( __AVR_ATtiny24__ ) || defined( __AVR_ATtiny44__ ) || defined( __AVR_ATtiny84__ )
+#if defined( __AVR_ATtiny24__ ) || defined( __AVR_ATtiny44__ ) || defined( __AVR_ATtiny84__ ) || defined( __AVR_ATtiny84A__)
 #define __AVR_ATtinyX4__
 #endif

sej7278:
if you look at the ide and tiny sourcecode, you can see -D is being passed to avrdude:

https://github.com/arduino/Arduino/blob/57dcf07244629b4492ad49fb7c1188694a2907ef/hardware/arduino/avr/platform.txt#L94

https://github.com/arduino/Arduino/blob/d8e5997328b76c570c1803baef4c5399783fded0/app/src/processing/app/debug/AvrdudeUploader.java#L168

https://code.google.com/p/arduino-tiny/source/browse/avr/platform.txt?repo=core1-0150#84

its certainly passing -D for an uno on 1.0.5 and 1.6.4, its even passing it for my mega2560 in fact (difference seems to be ```-c wiring``` instead of ```-c arduino```)

hhgarnes:
I use arduino-mk version: 1.3.4 from Ubuntu 14.10. I use the Arduino as an ISP, and I do make ispload. This sets -D around line 1072: AVRDUDE_OPTS = -q -V -D:
This is the file I have, I think: https://gitlab.mech.kuleuven.be/rob-ecs/arduino-car/raw/master/Arduino.mk

With the IDE I set the board as attiny84, choose Arduino as ISP as programmer, and I choose upload via ISP to program the chip. This does not set -D:

/usr/share/arduino/hardware/tools/avrdude -C/usr/share/arduino/hardware/tools/avrdude.conf -v -v -v -v -pattiny84 -cstk500v1 -P/dev/ttyACM0 -b19200 -Uflash:w:/tmp/build1348589659745212523.tmp/Blink.cpp.hex:i

sej7278:
well that must have been 1.3.4-1, before we changed it, it now does exactly the same as the IDE.

if you use arduino-mk [1.5-2](https://launchpad.net/ubuntu/wily/amd64/arduino-mk/1.5-2) (or git) you'll see that -D is only set in ```AVRDUDE_ARD_OPTS``` (for arduino's with USB) not in ```AVRDUDE_ISP_OPTS``` (for ISP'ing bare chips)

sej7278:
is the simpler patch working for you now?

sorry to be a pain but could you rebase it, 5 commits for 4 lines of code changes is a bit much ;-)

and maybe add something to HISTORY.md ?

```diff
diff --git a/HISTORY.md b/HISTORY.md
index 0930110..cc16346 100644
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -10,6 +10,7 @@ I tried to give credit whenever possible. If I have missed anyone, kindly add it
 - Tweak: Various spelling/grammar/typo fixes (https://github.com/dcousens)
 - Tweak: Clarified some 1.5+ issues in docs (Issue #352) (https://github.com/sej7278)
 - Tweak: Added some more Continuous Integration tests (https://github.com/sej7278)
+- New: Add option to split avrdude MCU from avr-gcc MCU (Issue #357) (https://github.com/hhgarnes)
 
 ### 1.5 (2015-04-07)
 - New: Add support for new 1.5.x library layout (Issue #275) (https://github.com/lukasz-e)
```

hhgarnes:
Ok, so that clears a couple of things up.

Sure I can re-base it. I have no idea how, though, as I'm basically crawling my way around git, but I'll get there in the end. I was just cleaning up a couple of things, and when I chose to commit directly into that patch-1 branch I figured the end commit would be a single diff.

sej7278:
ah ok, easiest way to do it is to copy your modified Arduino.mk somewhere, then create a new branch and copy it back into the new branch, or shall i do it?

sej7278:
argh, sorry, you need to add the changes to HISTORY.md in order, like i've put it above.

hhgarnes:
Ya, I'm fumbling around with the web interface. I'll try to make another fork, right after I fix the history.md.

sej7278:
i guess it needs the commits squashed with ```git rebase -i``` and then choose to squash all but one, but its complicated.

i think it would be easier to copy HISTORY.md and Arduino.mk into a new branch and do a single commit.

i hope this hasn't put you off git (or sending us pull requests!)

hhgarnes:
Nah, don't worry. I'm just not used to git, and I wasn't imagining the scope of this change as I set out doing a small patch in the web interface. I hope the newest pull request looks better to you. Can I delete pull requests somehow to make the noise go away?

sej7278:
you can just click the close button at the bottom of this page (and the other one).

i'm not keen on the web interface, it always seems easier to do it in a text editor on a local checkout

hhgarnes:
Crossing my fingers.

PR #359, opened 2015-06-01
Decouple the mcu flag between gcc and avrdude

hhgarnes:
Add support for decoupling the mcu flag between gcc and avrdude, which is for example useful if you want to compile with the mcu specified as attiny84a but upload with avrdude using mcu attiny84 (avrdude doesn't support such detailed chip specifications)

Also add some comments about interactions with attiny84a and avrdude's -D option, which can cause uploads to the chip through ArduinoISP to fail.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1282,13 +1282,22 @@ ifndef AVRDUDE_OPTS
     AVRDUDE_OPTS = -q -V
 endif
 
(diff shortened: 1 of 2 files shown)


sej7278:
looking good, travis-ci build should be fine :+1:

sudar:
@hhgarnes Thanks for the patch.

@sej7278 Thanks for following up and verifying the patch.

I just merged it after rebasing both the commits into a single one and tweaking the commit message.

PR #362, opened 2015-07-02
Added examples for various boards and IDE versions to the Blink example.

sej7278:
Apparently people are still finding it difficult to even get Blink working, largely due to 1.0 vs 1.6 confusion.

So I've added all the board/ide combinations i've used (nano, leonardo, uno, pro micro, teensy3, chipkit, mega2560, atmega1284p, atmega328p, attiny85, pinoccio, pro mini, fio, atmega-ng) as comments in the Blink example, and linked to it from README.md

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,6 +6,7 @@ I tried to give credit whenever possible. If I have missed anyone, kindly add it
 
 ### In Development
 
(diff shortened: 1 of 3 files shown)


sudar:
Thanks. This should be very helpful for beginners.

Issue #364, opened 2015-07-08
missing wiring_pulse.S from compilation


thinkberg:
MacOS X 10.10.4
avr stuff via homebrew
Arduino IDE 1.6.5

I have tried to compile a sketch which uses pulseIn(). This in turn uses a function from wiring_pulse.S that is not linked/compiled for some reason. The Arduino IDE 1.6.5 compiles it 

sources can be [found here](https://github.com/ubirch/ubirch-sketches)

Arduino IDE build process:
```
/Applications/Arduino.app/Contents/Java/hardware/tools/avr/bin/avr-gcc -c -g -x assembler-with-cpp -mmcu=atmega328p -DF_CPU=16000000L -DARDUINO=10605 -DARDUINO_AVR_UNO -DARDUINO_ARCH_AVR -I/Applications/Arduino.app/Contents/Java/hardware/arduino/avr/cores/arduino -I/Applications/Arduino.app/Contents/Java/hardware/arduino/avr/variants/standard /Applications/Arduino.app/Contents/Java/hardware/arduino/avr/cores/arduino/wiring_pulse.S -o /var/folders/c5/9sqmvxp54m3cmjb6hmtw2xlm0000gn/T/build5848034434382423610.tmp/wiring_pulse.S.o 
````

what my make produces (countPulseASM is in wiring_pulse.S which is not mentioned during the build process):
```
/Applications/Arduino.app/Contents/Java/hardware/tools/avr/bin/avr-gcc -mmcu=atmega328p -Wl,--gc-sections -Os -o build-uno/hc-sr04-led.elf build-uno/hc-sr04-led.o build-uno/libcore.a  -lc -lm
build-uno/libcore.a(wiring_pulse.o): In function `pulseIn':
wiring_pulse.c:(.text.pulseIn+0x32): undefined reference to `countPulseASM'
collect2: error: ld returned 1 exit status
make: *** [build-uno/hc-sr04-led.elf] Error 1
```

My makefile:

```
# http://hardwarefun.com/tutorials/compiling-arduino-sketches-using-makefile

#BOARD_TAG=uno
ARDUINO_PORT=/dev/cu.SLAB_USBtoUART

include /usr/local/opt/arduino-mk/Arduino.mk
```

From a quick look, the arduino-makefile looks at all the source types, maybe it misses on wiring_pulse because it exists as wiring_pulse.c and wiring_pulse.S?

sej7278:
[...]

PR #266, opened 2014-09-16
Preserve original source file extension in object file.

peplin:
See #255 for discussion

PR #365, opened 2015-07-09
Preserve source extension for object files.

sej7278:
Needed to fix the new wiring_pulse.S in IDE 1.6.5 which also has a wiring_pulse.c source file.

Mostly rebased @peplin's PR #266 and #256 so allows us to support newer chipKIT builds too.

See #364 for discussion.


sudar:
Is it good to merge now or do you want to wait till we do more testing for this?

sej7278:
i think we may as well merge it, i'm sure @peplin tested the PR's its based on previously on chipkit hardware

PR #369, opened 2015-07-14
Added support for AVR Dragon in both ISP and DebugWire mode

mtnocean:
Adds support for
`ISP_PROG = dragon_isp`
and
`ISP_PROG = dragon_dw`

ISP mode is reasonable to add, but I think there might be a question as to whether DebugWire mode is appropriate. There might be some confusion since the makefile command is `make ispload` while the programmer would be using DebugWire.

I hope I'm doing this right. This is my first pull request.
Thanks for all your work.


sej7278:
the dragon doesn't have a port in either mode, so makes sense to me :+1:

Issue #372, opened 2015-08-07
Allow spaces in project path.


MeanEYE:
If I save my project in `LCD Test` directory, `makefile` will fail to compile and upload. However if I rename directory to `LCD_Test` everything works properly.

sudar:
This Makefile doesn't support directories with spaces, since it breaks lot of scripts that the makefile uses.

Also it is generally a bad practice to have space in directory names anyways :) So would it be possible for you to change it?

MeanEYE:
Well, of course it's possible, but it's a bad excuse on their part. Should we limit file names to 8 characters now? Since we are already making an effort of living in the past. Makefile should respect specs of the system in question, and not invent their own rules. Even DOS supported spaces. Gah, I love when developers come up with retarded ideas.

Thanks anyway, I can see this is a wide spread problem and that I should probably start looking for another solution as `make` is not going to cut it anymore. Also, since this is invalid issue report related to this project, am closing it. Thanks for clearing this up.

sudar:
> I love when developers come up with retarded ideas.

Well, I can give you the reason why we don't currently support directories with spaces (or other special characters) and you can decide whether it is _retarded_ or not.

This Makefile uses lot of shell scripts and spaces (or other special characters) in the directory names breaks those shell scripts. It would takes us huge amount of effort to go over all those shell scripts and properly quote them and then test them in all the different operating systems that we support. Instead we choose to spend that time to add new features and gently (without calling them _retarded_) tell the users to have rename their directories.

Ohh.. by the way anyone is welcomed to help us the developers (who have come up with _retarded_ ideas) by coming up with a clever idea and/or a (well tested) pull request to add support for directories with spaces (or any other special characters)

ladislas:
@sudar :+1: 

@MeanEYE :-1:

MeanEYE:
I thought `make` tool itself was responsible for lack of support, but it looks like that's not the case. _Retarded_ might be too harsh of a word to use to describe this issue but telling people to adapt and not fix the issue because it's too much work is not a good policy in my opinion.

If you are, by any chance, offended by my statement, I welcome you to submit a patch to any open source project which removes support for spaces in paths and see how that goes and if they are willing to accept it. Should such a patch be accepted I am more than willing to publicly apologize for my choice of words. Until then I stand by my statement as we do live in 21st century which tends to have support for many things, including spaces in paths, regardless if it's a bad policy or not.

[...]

Issue #374, opened 2015-08-20
ATtiny85 creates a .elf file named "ATtiny.elf"


GreyGnome:
Hello,
When I compile a sketch for ATtiny85, the file it creates in the build directory is named "ATtiny.elf". The sketch is named "ATtinyBlink.ino" so I would hope it would create "ATtinyBlink.elf". I have a utility that I use to dump the object file and look at the assembler, and it works fine on the ATmega328 and 2560 but here it breaks because of the file name difference. Is it something I've done? Here's some of the gcc output which shows where it's creating the elf file:
```
/usr/bin/avr-gcc -mmcu=attiny85 -Wl,--gc-sections -Os -o build-attiny85/ATtiny.elf build-attiny85/ATtinyBlin
k.o build-attiny85/libcore.a  -lc -lm
/usr/bin/avr-objcopy -j .eeprom --set-section-flags=.eeprom='alloc,load' \
        --change-section-lma .eeprom=0 -O ihex build-attiny85/ATtiny.elf build-attiny85/ATtiny.eep
/usr/bin/avr-objcopy: --change-section-lma .eeprom=0x0000000000000000 never used
/usr/bin/avr-objcopy -O ihex -R .eeprom build-attiny85/ATtiny.elf build-attiny85/ATtiny.hex

/usr/bin/avr-size --mcu=attiny85 -C --format=avr build-attiny85/ATtiny.elf

```
Here is my Makefile:

```
# Arduino Make file. Refer to https://github.com/sudar/Arduino-Makefile

AVR_TOOLS_DIR = /usr
CFLAGS_STD = -g -I /home/schwager/.arduino15/packages/attiny/hardware/avr/1.0.1/variants/tiny8
CXXFLAGS_STD = -g -I /home/schwager/.arduino15/packages/attiny/hardware/avr/1.0.1/variants/tiny8

# if you have placed the alternate core in your sketchbook directory, then you can just mention the core name alone.
ALTERNATE_CORE = attiny
# If not, you might have to include the full path.
#ALTERNATE_CORE_PATH = /home/schwager/bin/arduino-1.6.4/hardware/arduino/avr/cores/
ALTERNATE_CORE_PATH = /home/schwager/.arduino15/packages/attiny/hardware/avr/1.0.1
ARDUINO_VAR_PATH = /home/schwager/Projects/Arduino/attiny-ide-1.6.x/attiny/avr/variants
[...]

PR #379, opened 2015-09-06
Make sure TARGET is set correctly when CURDIR contains spaces.

svendahlstrand:
With this fix the `TARGET` variable is set correctly when the project exists in
a path that contains spaces. So in this case:

    /Users/Joe/Dropbox (Personal)/example-project

`TARGET` will be set to `example-project` instead of `Dropbox example-project` (like it was
before this fix).

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -266,7 +266,9 @@ endif
 # Default TARGET to pwd (ex Daniele Vergini)
 
 ifndef TARGET
-    TARGET  = $(notdir $(CURDIR))
+    space :=
+    space +=
+    TARGET = $(notdir $(subst $(space),_,$(CURDIR)))
 endif
 
 ########################################################################
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -17,6 +17,7 @@ I tried to give credit whenever possible. If I have missed anyone, kindly add it
 - Tweak: Added some more Continuous Integration tests (https://github.com/sej7278)
 - Tweak: Updated Fedora instructions (https://github.com/sej7278)
 - Fix: Preserve original extension for object files, support asm sources in core, fixes pulseInASM (Issue #255, #364) (https://github.com/sej7278)
+- Fix: Make sure TARGET is set correctly when CURDIR contains spaces.
 
 ### 1.5 (2015-04-07)
 - New: Add support for new 1.5.x library layout (Issue #275) (https://github.com/lukasz-e)


sej7278:
Hi, thanks for the PR, i was hoping it would fix #372 but unfortunately it doesn't cover the use case where there are spaces in the ```TARGET``` directory, you end up with:

```
silly people/
└── use spaces
    └── in paths
        ├── blink.ino
        ├── build-leonardo
        │   ├── blink.ino.d
        │   ├── blink.ino.o
        │   ├── core
        │   │   ├── avr-libc
        │   │   │   ├── malloc.c.d
        │   │   │   ├── malloc.c.o
        │   │   │   ├── realloc.c.d
        │   │   │   └── realloc.c.o
        │   │   ├── CDC.cpp.d
        │   │   ├── CDC.cpp.o
        │   │   ├── ... blah ...
        │   │   └── WString.cpp.o
        │   ├── libcore.a
        │   ├── paths.eep
        │   ├── paths.elf
        │   ├── paths.hex
        │   └── paths.hex.sizeok
        └── Makefile
```

i.e. you get an elf file named after the ```lastword``` of the directory instead of the ino file, which may be way happened in #374 (if his directory was "something ATtiny").

Although as long as TARGET doesn't have a space, it seems ok - might be worth adding a note about that to #380.

svendahlstrand:
I try to avoid spaces in file and directory names myself, so I didn't think about the case when the base name contains spaces. You are absolutely right! :) I've fixed this pull request and updated #380 accordingly.

sej7278:
could we make it replace spaces with underscores to be a bit more "standard" and so we don't get confused with the dash in ```OBJDIR```?

so we'd get something like:

```
space man/
├── blink.ino
├── build-uno/
│   ├── blink.ino.d
│   ├── blink.ino.o
│   ├── core/
│   ├── libcore.a
│   ├── space_man.eep
│   ├── space_man.elf
│   ├── space_man.hex
│   └── space_man.hex.sizeok
└── Makefile
```

it would need the text in #380 updating too :-(

then i'd say we can merge. you ok with that @sudar ?

svendahlstrand:
I've updated the Makefile and documentation to use underscore.

I'm not sure why the directory name is used as default `TARGET`. Is a `ino` file required for everything to work or could one use a `c` or `cpp` file instead? If so, maybe thats why the directory name is used.

sej7278:
yes you can use c/cpp files not just ino, maybe that's why it uses the directory instead

Issue #381, opened 2015-09-08
AVRDUDE_CONF not properly set when AVR_TOOLS_DIR specified


theicfire:
Looking at `Arduino.mk`, I see that if you're not on Windows and have `AVR_TOOLS_DIR` specified, AVRDUDE_CONF will not be set.

Repro:
- Set `AVR_TOOLS_DIR`
- Run `make upload` not on windows

The final upload command is something like
`/Applications/Arduino.app/Contents/Java/hardware/tools/avr/bin/avrdude -q -V -p atmega328p -D -c arduino -b 57600 -P /dev/cu.usbserial-AL00EZ5A -U flash:w:build-pro-16MHzatmega328/Blink.hex:i`

But if I remove `AVR_TOOLS_DIR`, I get something like:
`/Applications/Arduino.app/Contents/Java/hardware/tools/avr/bin/avrdude -q -V -p atmega328p -C /Applications/Arduino.app/Contents/Java/hardware/tools/avr/etc/avrdude.conf -D -c arduino -b 57600 -P /dev/cu.usbserial-AL00EZ5A -U flash:w:build-pro-16MHzatmega328/Blink.hex:i`

This second one actually works in that it has avrdude.conf passed over. The first one does not.

sej7278:
whilst i agree that setting ```AVR_TOOLS_DIR``` on non-windows does seem to cause ```AVRDUDE_CONF``` not to be found, it seems the fix is over-simplistic as it totally breaks linux:

```
/usr/share/arduino/hardware/tools/avr/bin/avrdude -q -V -p atmega2560  -C /usr/share/arduino/hardware/tools/avr/etc/avrdude.conf
...
avrdude: can't open config file "/usr/share/arduino/hardware/tools/avr/etc/avrdude.conf": No such file or directory
avrdude: error reading system wide configuration file "/usr/share/arduino/hardware/tools/avr/etc/avrdude.conf"
```

on linux it resides in /usr/share/arduino/hardware/tools/avr/../avrdude.conf which is a symlink to /etc/avrdude.conf

theicfire:
Ah, good catch. Looks like the if statement then should be "if windows or mac". I'm just not sure how to do that! It looks surprisingly complicated in Makefiles. Here's what I could find:

http://stackoverflow.com/questions/7656425/makefile-ifeq-logical-or

sej7278:
i think platform needs to be ignored, and a different check put in place, as we also support freebsd.

[...]

PR #382, opened 2015-09-08
AVRDUDE_CONF gets set when not in windows

theicfire:
Relating to sudar/Arduino-Makefile#381


sej7278:
do not merge, see discussion (breaks linux)

Issue #386, opened 2015-09-14
VENDOR variable conflicts with tcsh


stefanrueger:
Arduino.mk has stopped working under tcsh. It would roll out numerous grep errors of the kind

  grep: /usr/share/arduino-m/hardware/unknown/avr/boards.txt: No such file or directory

It took me a while to figure out that the problem is that VENDOR is a default environment name in the tcsh, which is wrongly taken to be a USER set value by Arduino.mk and makes it look in the wrong directories.

There are a number of things that can be done to improve this situation (also for bash users):
 - Test for the existence of the directory and make the error message more meaningful
 - Explain workarounds (set VENDOR explicitly in your Makefile, unsetenv VENDOR in tcsh, use bash, ...)
 - Rename the VENDOR variable in Makefile to something else, eg, ARDUINO_VENDOR, ARDMK_VENDOR, etc.

Obviously, I prefer the latter, as this avoids giving unwanted trouble to unsuspecting tcsh users, and descriptive variable names are always neat.

sej7278:
I've created PR #388 for you to test

stefanrueger:
That's cool. I can confirm this version works like a charm.

PR #388, opened 2015-09-14
Replace $VENDOR with $ARDMK_VENDOR

sej7278:
....as ```VENDOR``` is a tcsh environment variable.

```ARCHITECTURE``` is probably safe as that's usually called ```ARCH```

Fixes issue #386.

Need to decide if this is going to upset too many user's who have already started using ```VENDOR``` - and who uses tcsh? ;-)

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -319,11 +319,11 @@ endif
 
 ########################################################################
 # 1.5.x vendor - defaults to arduino
-ifndef VENDOR
-	VENDOR = arduino
-    $(call show_config_variable,VENDOR,[DEFAULT])
+ifndef ARDMK_VENDOR
+	ARDMK_VENDOR = arduino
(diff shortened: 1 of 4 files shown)


stefanrueger:
That's col. I can confirm that works like a charm.

ladislas:
I confirm it works on OS X on my projects as well

sej7278:
thanks @ladislas

sudar:
@sej7278 Works for me too.

I think since we introduced `VENDOR` only in 1.5.x we can get away with not being backward compatible.

Also @sej7278 Thanks for taking care of the recent couple of tickets :)

sej7278:
yes i came to the same conclusion regarding it being a 1.5 variable.

do you want to do the honours then @sudar ?

sudar:
@sej7278 

Sure :) Just merged it. Thanks.

PR #390, opened 2015-09-21
Correct path ARDUINO_DIR for Macs

thomaskilian:
Hi, as posted on your blog I send you a correction for the wrong Mac path.
Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ On the Mac with IDE 1.0 you might want to set:
 
 On the Mac with IDE 1.5+ it's like above but with
 
+```
     ARDUINO_DIR   = /Applications/Arduino.app/Contents/Java
 ```
 
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ whilst [Blink](examples/Blink/Makefile) demonstrates the minimal settings requir
 
 Download a copy of this repo some where in your system or install it through a package.
 
-On the Mac you might want to set:
+On the Mac with IDE 1.0 you might want to set:
 
 ```make
     ARDUINO_DIR   = /Applications/Arduino.app/Contents/Resources/Java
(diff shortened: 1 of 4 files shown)


sej7278:
its not wrong, its just that you're using 1.6 not 1.0

thomaskilian:
Hmm, Maybe then it should contain both paths?

sej7278:
you mean like this:

https://github.com/sudar/Arduino-Makefile/blob/master/arduino-mk-vars.md#arduino_dir

:smile:

thomaskilian:
Exactly :-)

sej7278:
its only an example, and personally I think Blink is better, there's more info in there, although not for osx i guess.

https://github.com/sudar/Arduino-Makefile/blob/master/examples/Blink/Makefile

if you want to modify your PR with some explanatory text about the two versions, i'm sure we can merge it, thanks.

thomaskilian:
Ok. I have changed all places I could find. Will send you a new PR

thomaskilian:
Sorry, I'm new to this mechanism. Looks like my updates are already forwarded?

sej7278:
yes that's how it works, you don't need a new PR, it just munges together all your commits.

it looks like you've missed the opening triple-backticks ``` at line 148 as the "On Linux...." bit is still in the code tags.

could you also add something to HISTORY.md for example:

```
- Fix: Document OSX 1.0/1.6 ARDUINO_DIR differences (https://github.com/thomaskilian)
```

README.md
```diff
@@ -144,6 +144,11 @@ On the Mac you might want to set:
     BOARD_TAG     = mega2560
 ```
 
+On the Mac with IDE 1.5+ it's like above but with
+
```
sej7278: needs backticks here ```

thomaskilian:
Thanks, will do so.

Issue #389, opened 2015-09-19
Adafruit Bluefruit LE Micro uploading fails


jonblack:
The [Adafruit Bluefruit LE Micro](https://www.adafruit.com/products/2661) is an Arduino compatible board which contains a Bluetooth LE chip and an ATmega32u4.

The documentation says the following:

>You can program the ATmega32u4 over USB using the built-in USB bootloader, either directly with avrdude or using the Arduino IDE (it's the same as used as in the Adafruit Flora or Arduino Leonardo).

When selecting `Leonardo` in the Arduino IDE, compilation and uploading of the `blink` example works fine. Using a Makefile fails to upload with the following error:

    mkdir -p build-leonardo
    make reset
    make[1]: Entering directory '/home/jon/arduino/example'
    /usr/bin/ard-reset-arduino --caterina  /dev/ttyACM0
    Traceback (most recent call last):
      File "/usr/bin/ard-reset-arduino", line 23, in <module>
        ser.setDTR(False) # (see Arduino IDE source code)
      File "/usr/lib/python2.7/dist-packages/serial/serialposix.py", line 564, in setDTR
        fcntl.ioctl(self.fd, TIOCMBIC, TIOCM_DTR_str)
    IOError: [Errno 32] Broken pipe
    /usr/share/arduino/Arduino.mk:1432: recipe for target 'reset' failed
    make[1]: *** [reset] Error 1
    make[1]: Leaving directory '/home/jon/arduino/example'
    /usr/share/arduino/Arduino.mk:1408: recipe for target 'upload' failed
    make: *** [upload] Error 2

The Makefile contains:

    BOARD_TAG = leonardo
    ARDUINO_SKETCHBOOK = /home/jon/arduino
    include /usr/share/arduino/Arduino.mk

I'm using version 1.6.5 of the Arduino IDE and am on Debian.

sej7278:
i saw this recently too, i think it was a busted 32u4 bootloader on my pro micro, that was using 1.0.5 on debian

jonblack:
I actually uploaded the blink sketch successfully. All subsequence uploads produce the error. Do you still think it could be a wiring issue? In any case, I'm getting the same problems in the Arduino IDE so it's not a problem with this project.

ladislas:
have you tried manually resetting the board just before uploading?

sej7278:
i assume its not the issue on their product page?

[...]

PR #393, opened 2015-10-26
Better parsing of boards.txt file

zabereer:
The Adafruit boards.txt file contains trinket3, and protrinket3 which are both matched by the existing boards.txt parsing when specifying BOARD_TAG=trinket3.
This change anchors the grep regex matching to beginning of the line.
Thank you for a great Arduino makefile.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -6,12 +6,12 @@ I tried to give credit whenever possible. If I have missed anyone, kindly add it
 
 ### In Development
 
-- Tweak: Add support for Adafruit trinket3/trinket5/protrinket3/protrinket5 by improved BOARDS_TXT parsing
 - New: Add AVR Dragon to list of ISP's without a port (https://github.com/mtnocean)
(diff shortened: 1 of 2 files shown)


sej7278:
good spot! :+1: 

i can't believe we didn't notice it whilst working on #389 but i guess we didn't select one of the boards whose name was a subset of another board (bad adafruit, naughty!)

HISTORY.md
```diff
@@ -6,6 +6,7 @@ I tried to give credit whenever possible. If I have missed anyone, kindly add it
 
 ### In Development
 
+- Tweak: Add support for Adafruit trinket3/trinket5/protrinket3/protrinket5 by improved BOARDS_TXT parsing
```
sej7278: can we move this down to where the other Tweak lines are? also credit yourself in brackets (https://github.com/zabereer)

zabereer: Sure, I added a commit to fix the HISTORY.md.
Good news you can't break it :-)
Let me know if you prefer a collapsed single commit than the two.

PR #412, opened 2016-02-19
Added target to display submenu options parsed out of boards.txt

drewhutchison:
Added a make target `show_submenu` to aid in setting `BOARD_SUB` analogously to using `show_boards` for `BOARD_TAG`.


sudar:
Thanks. Just merged it.

Issue #29, opened 2012-12-05
Arduino.mk insists on building way too many unnecessary libraries


kenrestivo:
I have this simple makefile 

``` makefile
ARDUINO_DIR   = /usr/share/arduino
ARDMK_DIR     = /home/localkens/embed/src/Arduino-Makefile
AVR_TOOLS_DIR = /usr

BOARD_TAG    = diecimila
ARDUINO_PORT = /dev/ttyUSB4

ARDUINO_LIBS= SoftwareSerial

ARDUINO_USER_LIBS = ModbusMaster

USER_LIB_PATH=../../../

include /home/localkens/embed/src/Arduino-Makefile/arduino-mk/Arduino.mk
```

And make seems to insist on building all kinds of insane stuff, like HID and USB, and IPAddress???!, when I did not tell it to do that.

```
/usr/bin/avr-gcc -c -mmcu=atmega168 -DF_CPU=16000000L -DARDUINO=100 -I. -I/usr/share/arduino/hardware/arduino/cores/arduino -I/usr/share/arduino/hardware/arduino/variants/standard -I/usr/share/arduino/libraries/SoftwareSerial -I../../..//ModbusMaster -g -Os -w -Wall -DUSB_VID= -DUSB_PID= -ffunction-sections -fdata-sections -std=gnu99 ../../..//ModbusMaster/ModbusMaster.cpp -o build-diecimila/libs/ModbusMaster/ModbusMaster.o
/usr/bin/avr-ar rcs build-diecimila/libcore.a  build-diecimila/WInterrupts.o  build-diecimila/wiring_analog.o  build-diecimila/wiring.o  build-diecimila/wiring_digital.o  build-diecimila/wiring_pulse.o  build-diecimila/wiring_shift.o  build-diecimila/CDC.o  build-diecimila/HardwareSerial.o  build-diecimila/HID.o  build-diecimila/IPAddress.o  build-diecimila/main.o  build-diecimila/new.o  build-diecimila/Print.o  build-diecimila/Stream.o  build-diecimila/Tone.o  build-diecimila/USBCore.o  build-diecimila/WMath.o  build-diecimila/WString.o  build-diecimila/libs/SoftwareSerial/SoftwareSerial.o build-diecimila/libs/ModbusMaster/ModbusMaster.o 
[...]

PR #415, opened 2016-02-29
Update Arduino.mk

thalesmaoa:
Added some break lines.

Just want to compile the program without arduino lib, no no_core flag does not let know processor size.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -697,6 +697,13 @@ ifeq ($(strip $(NO_CORE)),)
     endif
 
 endif
+ 
+ifndef HEX_MAXIMUM_SIZE
+        HEX_MAXIMUM_SIZE := $(call PARSE_BOARD,$(BOARD_TAG),menu.cpu.$(BOARD_SUB).upload.maximum_size)
+        ifndef HEX_MAXIMUM_SIZE
+            HEX_MAXIMUM_SIZE := $(call PARSE_BOARD,$(BOARD_TAG),upload.maximum_size)
+        endif
+endif
 
 # Everything gets built in here (include BOARD_TAG now)
 ifndef OBJDIR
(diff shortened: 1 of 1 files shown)


sudar:
@sej7278 What are you thoughts on this?

@thalesmaoa Can you also please update HISTORY.md file with a brief summary of the change?

sej7278:
i've no idea what its supposed to do lol

thalesmaoa:
I don't know really if it is the best way to fix a bug, but I will explain
what happen and you can figure if it is a good merge.


I've made a standalone program to use with an arduino based processor. When
I use your awesome makefile, it work ok, but it also compile arduino libs,
but I don't use them. I do not even include them.

I realized that no_core = yes and mcu = atmega2560 flag keeps arduino lib
out of compilation, but at the end throw a bug message.

This message is what I was trying to fix.

1) Just added breadlines to the end of echo

2) When I use mcu = 2560 it already knows my processor size, because it
still tells me how full I am. In my case, 18%. But in the end, it tells me
that it doesn't. So I took out HEX_MAXIMUM_SIZE variable so the message
doesn't start.

Best regards
​

sej7278:
seems odd though, as we put all of the calls to PARSE_BOARD inside the NO_CORE block as if you're not using the arduino core, you also wouldn't have a boards.txt file to parse.

surely the solution to your problem would be to put ```HEX_MAXIMUM_SIZE=258048``` in your Makefile?

also you've duplicated that code block, rather than moving it.

thalesmaoa:
You are right, but I do use arduino core. It was just my workaround to no
compile arduino libs.

Sorry to duplicate, I didn't notice, but it is not a good solution indeed.

-
Thales Alexandre Carvalho Maia


2016-03-01 8:50 GMT-03:00 Simon John <notifications@github.com>:

sej7278:
so you're using the arduino core but none of the libraries, and want to not compile the built-in libraries?

that's issue #29 which is the same way the ide does it - the libraries are compiled so its a bit slow, but they're not linked so take up no memory in the final hex.

thalesmaoa:
That is the reason why I quit IDE as well. :)

I know it does not increase hex size, but pollute my make output and
difficult the tracking of warning messages.
It is not a big deal, just a simple solution.

-
Thales Alexandre Carvalho Maia


2016-03-01 9:47 GMT-03:00 Simon John <notifications@github.com>:

sej7278:
well if you don't mind it compiling the unused libraries and just want it to be quiet in the output then probably setting ```HEX_MAXIMUM_SIZE=258048``` is the best way, rather than your PR which could cause problems for people who really don't want to use the arduino core or boards.txt at all.

[ARDUINO_QUIET](https://github.com/sudar/Arduino-Makefile/blob/master/arduino-mk-vars.md#arduino_quiet) may be of some interest to you also

sudar:
@thalesmaoa 

Thanks for the PR.

Please see @sej7278's suggestion above.

I am closing this without merging. But feel free to re-open if you have additional information.

Issue #433, opened 2016-06-18
Documentation: Python or perl


az-z:
hello,
in the guide http://hardwarefun.com/tutorials/compiling-arduino-sketches-using-makefile you state : "
Install dependencies

The Makefile delegates resetting the board to a short Perl program. You’ll need to install Device::SerialPort and YAML library."

in the README you state:
 "
pySerial
 The Makefile also delegates resetting the board to a short Python program. You'll need to install pySerial to use it though."

sudar:
You have to install pySerial. The documentation in github is the current.

The guide at  http://hardwarefun.com/tutorials/compiling-arduino-sketches-using-makefile is out of date. I will update it.

sej7278:
that's just an old blog post, its not the guide.

the guide is https://github.com/sudar/Arduino-Makefile#requirements and states its python not perl

az-z:
thank you gents. Then, if i may suggest,  update the home page (https://github.com/sudar/Arduino-Makefile) and either remove the reference to the blog or (better) just state that documentation supersedes the blog. The way i read it is - "go to the blog for the detailed how-to instructions".
Thank you

az-z:
let me figure out what i need to do and i will submit the pull request.

PR #434, opened 2016-06-18
Documentation update

az-z:
As the followup on ticket #433 .
Updated the documentation to reflect current state of Linux implementation.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -27,18 +27,18 @@
 #
 # PATHS YOU NEED TO SET UP
 #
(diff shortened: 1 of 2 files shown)


az-z:
i messed up the markup on the README. will resend the request

sej7278:
it would be nice to be able to preview markup without committing it, plenty of crap websites try it, but would expect github to be able to do something about it.

az-z:
Especially when the changes somewhere in the middle of the document. ..
pita..

PR #445, opened 2016-07-29
Fix for 1.6.10 version detection

### LINKED ISSUE #444: 1.6.10 breaks version detection
1.6.10 currently returns ```ARDUINO_VERSION = 161```

This patch seems to work for me - even with Debian-esque "105+dfsg2-4" versions

Tested on 1.0.5 (debian) and 1.0.6, 1.6.9, 1.6.10 from arduino.cc linux64

Any comments - I've replaced the zero-padding with "delete the plus sign and anything after it" and we retrieve the last 4 bytes instead of 3, but that doesn't seem to break when there are only 3 bytes to return (e.g. 169)

Its only pedantic I guess anyway, as we only ever test for > 150 

```diff
diff --git a/Arduino.mk b/Arduino.mk
index ffda348..6b2741f 100644
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -283,7 +283,7 @@ ifndef ARDUINO_VERSION
     # Remove all the decimals, and right-pad with zeros, and finally grab the first 3 bytes.
     # Works for 1.0 and 1.0.1
     VERSION_FILE := $(ARDUINO_DIR)/lib/version.txt
-    AUTO_ARDUINO_VERSION := $(shell [ -e $(VERSION_FILE) ] && cat $(VERSION_FILE) | sed -e 's/^[0-9]://g' -e 's/[.]//g' -e 's/$$/0000/' | head -c3)
+    AUTO_ARDUINO_VERSION := $(shell [ -e $(VERSION_FILE) ] && cat $(VERSION_FILE) | sed -e 's/^[0-9]://g' -e 's/[.]//g' -e 's/\+.*//g' | head -c4)
     ifdef AUTO_ARDUINO_VERSION
         ARDUINO_VERSION = $(AUTO_ARDUINO_VERSION)
         $(call show_config_variable,ARDUINO_VERSION,[AUTODETECTED])
```

sej7278:
Fixes #444 

Also updated HISTORY.md with latest commits

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -280,10 +280,10 @@ endif
 # Arduino version number
 
 ifndef ARDUINO_VERSION
-    # Remove all the decimals, and right-pad with zeros, and finally grab the first 3 bytes.
-    # Works for 1.0 and 1.0.1
+    # Remove all the decimals, remove anything before/including ":", remove anything after/including "+" and finally grab the last 5 bytes.
+    # Works for 1.0 and 1.0.1 and 1.6.10 and debian-style 2:1.0.5+dfsg2-4
     VERSION_FILE := $(ARDUINO_DIR)/lib/version.txt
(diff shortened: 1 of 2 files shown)


sudar:
As always thanks @sej7278

Issue #446, opened 2016-07-29
Look into new LTO (etc.) flags


sej7278:
All sorts of new compiler flags seem to have been added in 1.6.10, probably need to compare to ours, figure out what they do and think about merging, along the lines of #442, initially it looks like this lot:

CXXFLAGS: ``` -flto -w -x c++ -E -CC```
CFLAGS: ```-std=gnu11 -flto -fno-fat-lto-objects```

  -E (preprocess only; do not compile, assemble or link)
  -x c++ (specify language is c++)

no idea what -w (lowercase not upper) or -CC do

sej7278:
Discussion https://github.com/arduino/Arduino/issues/660

sej7278:
Fixed in PR #452

PR #451, opened 2016-09-08
Added LTO flags

sej7278:
LTO functionality was added in IDE 1.6.10 to provide smaller sizes and faster code (and apparently build times). See issue #446

Will also work with ide 1.0.x as long as you have avr-gcc toolchain > 4.8 (e.g. debian/ubuntu)

On a nano328 for example:

Blink without LTO patch (HEAD):
```
Program:    1112 bytes (3.4% Full)
(.text + .data + .bootloader)

Data:         11 bytes (0.5% Full)
(.data + .bss + .noinit)
```

Blink with LTO patch:
```
Program:     946 bytes (2.9% Full)
(.text + .data + .bootloader)

Data:          9 bytes (0.4% Full)
(.data + .bss + .noinit)
```

This is only relevant to AVR.


sej7278:
don't merge this yet as it seems the version check should probably be against avr-gcc 4.8 or newer, not ide 1.5 or newer, potentially these improvements can be used with ide 1.0 on e.g. debian which uses avr-gcc 4.9.2

Issue #456, opened 2016-09-18
program has been built without plugin support


SirRibeiro:
I get this error when a try to execute make ( sorry - this program has been built without plugin support ) :

```
/git/Arduino-Makefile/examples/Blink$ make
-------------------------
Arduino.mk Configuration:
- [AUTODETECTED]       CURRENT_OS = LINUX 
- [USER]               ARDUINO_DIR = /home/code/arduino 
- [USER]               ARDMK_DIR = /home/code/git/Arduino-Makefile 
- [AUTODETECTED]       ARDUINO_VERSION = 1710 
- [DEFAULT]            ARCHITECTURE = avr 
- [DEFAULT]            ARDMK_VENDOR = arduino 
- [AUTODETECTED]       ARDUINO_PREFERENCES_PATH = /home/code/.arduino15/preferences.txt 
- [AUTODETECTED]       ARDUINO_SKETCHBOOK = /home/code/Arduino (from arduino preferences file)
- [USER]               AVR_TOOLS_DIR = /home/code/arduino/hardware/tools/avr 
- [COMPUTED]           ARDUINO_LIB_PATH = /home/code/arduino/libraries (from ARDUINO_DIR)
- [COMPUTED]           ARDUINO_PLATFORM_LIB_PATH = /home/code/arduino/hardware/arduino/avr/libraries (from ARDUINO_DIR)
- [COMPUTED]           ARDUINO_VAR_PATH = /home/code/arduino/hardware/arduino/avr/variants (from ARDUINO_DIR)
- [COMPUTED]           BOARDS_TXT = /home/code/arduino/hardware/arduino/avr/boards.txt (from ARDUINO_DIR)
- [DEFAULT]            USER_LIB_PATH = /home/code/Arduino/libraries (in user sketchbook)
- [DEFAULT]            PRE_BUILD_HOOK = pre-build-hook.sh 
- [USER]               BOARD_TAG = uno 
- [COMPUTED]           CORE = arduino (from build.core)
- [COMPUTED]           VARIANT = standard (from build.variant)
- [COMPUTED]           OBJDIR = build-uno (from BOARD_TAG)
- [COMPUTED]           ARDUINO_CORE_PATH = /home/code/arduino/hardware/arduino/avr/cores/arduino (from ARDUINO_DIR, BOARD_TAG and boards.txt)
- [ASSUMED]            MONITOR_BAUDRATE = 9600 
[...]

PR #464, opened 2016-10-11
Changed LTO support check to avr-gcc > 4.9.0 

sej7278:
The avr-gcc 4.8.1 bundled with the IDE prior to 1.6.10 doesn't support LTO or plugins. Fixes Issue #456 

So essentially LTO support will only be enabled with avr-gcc 4.9.2 which comes with 1.6.10 or later
and Debian, Ubuntu etc; not 4.8.1 which comes with IDE 1.6.9 and a few earlier versions.

Tested with:

 * 1.6.8 (avr-gcc 4.8.1 which doesn't support LTO so uses avr-ar and doesn't set LTO flags)
 * 1.6.12 (avr-gcc 4.9.2 which supports LTO so uses avr-gcc-ar and sets LTO flags)
 * 1.0.5 with Debian avr-gcc 4.9.2 (supports LTO so uses avr-gcc-ar and sets LTO flags)

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1032,7 +1032,7 @@ CC_VERNUM = $(shell $(CC) -dumpversion | sed 's/\.//g')
 
 # moved from above so we can find version-dependant ar
 ifndef AR_NAME
-    ifeq ($(shell expr $(CC_VERNUM) '>' 480), 1)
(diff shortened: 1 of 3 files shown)


sudar:
@sej7278 As always, thanks for taking care of this.

Issue #461, opened 2016-10-03
build.variant from boards.txt file is not parsed properly


imranmunshi:
I was trying to build the edtracker 9150 code using their custom core. The boards.txt has edtracker9150.build.variant=edtracker:edtracker2-9150

But that is not parsed properly by the makefile due to which I have to add VARIANT manually in the project Makefile. The Arduino.mk parses it to produce edtracker:edtracker2-9150 which actually should just be edtracker2-9150

sej7278:
looks like the edtracker is using a subdirectory for some reason

imranmunshi:
Arduino IDE 1.6.11-2 (arduino-bin PKGBUILD from AUR, Arch Linux) parses it properly and gives no issues.

sej7278:
can you give us a link as to where you got the edtracker core, also the directory structure within your hardware directory would be useful i.e. is it ~/sketchbook/hardware/edtracker/edtracker2-9150/ ?

the only core i could find uses ~/sketchbook/hardware/edtracker/avr/variants/edtracker2/ which is totally wrong

sej7278:
this seems to work for me - basically if there is a colon, set $VARIANT to the value on the right of the colon. doesn't break cores without colons e.g. arduino 1.0.5 with an uno:

```diff
diff --git a/Arduino.mk b/Arduino.mk
index 133a9a2..54f4920 100644
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -587,9 +587,9 @@ ifeq ($(strip $(NO_CORE)),)
 
     # Which variant ? This affects the include path
     ifndef VARIANT
-        VARIANT := $(call PARSE_BOARD,$(BOARD_TAG),menu.(chip|cpu).$(BOARD_SUB).build.variant)
+        VARIANT := $(shell echo $(call PARSE_BOARD,$(BOARD_TAG),menu.(chip|cpu).$(BOARD_SUB).build.variant) | cut -d : -f 2)
         ifndef VARIANT
-            VARIANT := $(call PARSE_BOARD,$(BOARD_TAG),build.variant)
+            VARIANT = $(shell echo $(call PARSE_BOARD,$(BOARD_TAG),build.variant) | cut -d : -f 2)
         endif
         $(call show_config_variable,VARIANT,[COMPUTED],(from build.variant))
     else
```

sej7278:
I think changing PARSE_BOARD is better:

```diff
diff --git a/Arduino.mk b/Arduino.mk
[...]

PR #465, opened 2016-10-11
Changed the PARSE_BOARD function to handle colons

sej7278:
...found in things like build.core and build.variant

Fixes Issue #461


sudar:
Thanks

PR #469, opened 2016-11-12
Fix parsing of Teensy options with multiple "="s

wilywampa:
The new Teensy 3.5/3.6 boards have arrived and boards.txt now has some configuration lines that have multiple "=" signs that aren't handled correctly by the current value of `PARSE_TEENSY`. Specifically, `teensy35.build.flags.cpu` get cuts off at the second "=" the the `-mfloat-abi` option has no value and the rest of the options are left out. Fortunately it's simple to fix by changing the field option of the cut command.

### DIFF
--- a/Teensy.mk
+++ b/Teensy.mk
@@ -43,7 +43,7 @@ endif
 
 ifndef PARSE_TEENSY
     # result = $(call READ_BOARD_TXT, 'boardname', 'parameter')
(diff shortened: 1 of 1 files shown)


sej7278:
:+1: seems to work without breaking anything else in boards.txt

do wish they'd stop doing all this weird stuff with teensy cores!

sudar:
Thanks @wilywampa and @sej7278

PR #475, opened 2017-01-21
Fix generate_assembly for cpp files.

Batchyx:
Do not include the Arduino header when generating assembly for .cpp
files with generate_assembly.  This was likely a copy-paste error.

Also included is a commit to re-create a "In Development" section in HISTORY.md

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1262,7 +1262,7 @@ $(OBJDIR)/%.s: %.ino $(COMMON_DEPS) | $(OBJDIR)
 
 $(OBJDIR)/%.s: %.cpp $(COMMON_DEPS) | $(OBJDIR)
 	@$(MKDIR) $(dir $@)
-	$(CXX) -x c++ -include $(ARDUINO_HEADER) -MMD -S -fverbose-asm $(CPPFLAGS) $(CXXFLAGS) $< -o $@
+	$(CXX) -x c++ -MMD -S -fverbose-asm $(CPPFLAGS) $(CXXFLAGS) $< -o $@
 
 # core files
 $(OBJDIR)/core/%.c.o: $(ARDUINO_CORE_PATH)/%.c $(COMMON_DEPS) | $(OBJDIR)
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -4,6 +4,8 @@ A Makefile for Arduino Sketches
 The following is the rough list of changes that went into different versions.
 I tried to give credit whenever possible. If I have missed anyone, kindly add it to the list.
 
+### In Development
+
 ### 1.5.2 (2017-01-11)
 
 - New: Add LTO support for users with avr-gcc > 4.9 (issue #446 & #456) (https://github.com/sej7278)
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,6 +5,7 @@ The following is the rough list of changes that went into different versions.
 I tried to give credit whenever possible. If I have missed anyone, kindly add it to the list.
 
 ### In Development
+- Fix: Do not include the Arduino header when calling generate_assembly on .cpp files. (https://github.com/Batchyx)
 
 ### 1.5.2 (2017-01-11)
 


sudar:
@sej7278 

sej7278:
i don't understand it really, need more info

Batchyx:
The documentation mentions a target called "generate_assembly", which should run the compiler, but not the assembler, and should emit an assembly file instead of an object file.
```
make generate_assembly - generate a .s file containing the compiler\n\
                           generated assembly of the main sketch.\n\
```
It is implemented as follows:
```
generate_assembly: $(OBJDIR)/$(TARGET).s
		@$(ECHO) "Compiler-generated assembly for the main input source has been dumped to $(OBJDIR)/$(TARGET).s\n\n"
```

In order for the assembly to match what would have been compiled, the compiler should be invoked in the same way with the same options, but with the -S option instead of the -c option.  This is currently true for .ino and .pde files, but not for .cpp files, where the compiler command for compiling is:
```
$(OBJDIR)/%.cpp.o: %.cpp $(COMMON_DEPS) | $(OBJDIR)
	@$(MKDIR) $(dir $@)
	$(CXX) -MMD -c $(CPPFLAGS) $(CXXFLAGS) $< -o $@
```
while the compiler command to generate the compiler-generated assembly is:
```
$(OBJDIR)/%.s: %.cpp $(COMMON_DEPS) | $(OBJDIR)
	@$(MKDIR) $(dir $@)
	$(CXX) -x c++ -include $(ARDUINO_HEADER) -MMD -S -fverbose-asm $(CPPFLAGS) $(CXXFLAGS) $< -o $@
```

This difference was enough for my code to work when compiling, but fail when trying to generate assembly with `make generate_assembly`.  My code was something like this:
```
#include "my_header.h" // This defines a method called bit()
#include <Arduino.h> // Ick, This defines a macro called bit()
#undef bit // remove that macro
// ...
```

When trying to generate assembly, the -include $(ARDUINO_HEADER) forces an include of Arduino.h at the very beginning, and the compilation fails at `my_header.h` near the definition of my `bit()` method.

One may also want to use .cpp files with no cores and without using the Arduino.h header.  `make generate_assembly` would also fails there.

My patch simply removes the `-include $(ARDUINO_HEADER)` when generating the assembly for a .cpp file, to match the normal compilation command, which does not include it.

sej7278:
oh i've got you - so we have at line 1232 onwards the compiler commands:

```make
$(OBJDIR)/%.cpp.o: %.cpp $(COMMON_DEPS) | $(OBJDIR)
	@$(MKDIR) $(dir $@)
	$(CXX) -MMD -c $(CPPFLAGS) $(CXXFLAGS) $< -o $@
...
$(OBJDIR)/%.ino.o: %.ino $(COMMON_DEPS) | $(OBJDIR)
	@$(MKDIR) $(dir $@)
	$(CXX) -x c++ -include $(ARDUINO_HEADER) -MMD -c $(CPPFLAGS) $(CXXFLAGS) $< -o $@
```

but then to generate the assembly we have this starting at 1259, which should only include the header for *.ino files not *.cpp ones, to match the above:

```make
$(OBJDIR)/%.s: %.ino $(COMMON_DEPS) | $(OBJDIR)
	@$(MKDIR) $(dir $@)
	$(CXX) -x c++ -include $(ARDUINO_HEADER) -MMD -S -fverbose-asm $(CPPFLAGS) $(CXXFLAGS) $< -o $@

$(OBJDIR)/%.s: %.cpp $(COMMON_DEPS) | $(OBJDIR)
	@$(MKDIR) $(dir $@)
	$(CXX) -x c++ -include $(ARDUINO_HEADER) -MMD -S -fverbose-asm $(CPPFLAGS) $(CXXFLAGS) $< -o $@
```

and we don't include the header for *.cpp files as the user should be doing that manually as per line 792.

so probably a copy'n'paste error. ok to merge from me :+1:

sudar:
Thanks @Batchyx and @sej7278

PR #477, opened 2017-02-04
Auto-detect F_CPU for Teensy boards

DaWelter:
Hello,
I am new user of your wonderful make scripts. Thanks a lot to all contributors. However when I wanted to try the blink example with my TeensyLC, I realized that the MCU frequency is apparently hardcoded to 96 MHz. Too much for this board which is supposed to run at 48 MHz.
Therefore I propose this little patch with which it should be possible to auto detect the fastest possible speed that can be found in boards.txt.
Went through all Teensy variants. The script selects F_CPU as expected for each of them.

### DIFF
--- a/Teensy.mk
+++ b/Teensy.mk
@@ -37,13 +37,22 @@ ARDMK_VENDOR        = teensy
 ARDUINO_CORE_PATH   = $(ARDUINO_DIR)/hardware/teensy/avr/cores/teensy3
 BOARDS_TXT          = $(ARDUINO_DIR)/hardware/$(ARDMK_VENDOR)/avr/boards.txt
 
-ifndef F_CPU
-    F_CPU=96000000
-endif
 
 ifndef PARSE_TEENSY
     # result = $(call READ_BOARD_TXT, 'boardname', 'parameter')
     PARSE_TEENSY = $(shell grep -v "^\#" "$(BOARDS_TXT)" | grep $(1).$(2) | cut -d = -f 2- )
+endif
+
+
+ifndef F_CPU
+  ifndef BOARD_SUB
+    SPEEDS := $(call PARSE_TEENSY,"$(BOARD_TAG),menu.speed.*.build.fcpu") # Obtain sequence of supported frequencies.
+    SPEEDS := $(shell printf "%d\n" $(SPEEDS) | sort -g) # Sort it, just in case. Printf to re-append newlines so that sort works.
+    F_CPU := $(lastword $(SPEEDS)) # List is sorted in ascending order. Take the fastest speed.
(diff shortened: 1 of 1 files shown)


sej7278:
i think i prefer the commented out line 52 solution, which uses BOARD_SUB rather than just selecting the fastest one

DaWelter:
I suggest a conditional so that the selection is automatic if BOARD_SUB ist not provided. Okay?
Meanwhile I updated by branch. Forgot to add -g to sort ...

sej7278:
sounds good to me

DaWelter:
Okay. Please take a look again.

In addition, I changed the parse regex for the fcpu spec so that it only produces matches at the beginning a line. A crude and simple way to deal with comments in board.txt.

Added quotation marks in the invocation of the teensy loader were required to handle project paths with spaces in it.

sej7278:
the fix looks good mostly, but we (well, make) don't support spaces in paths, we've been through this before and they cause more hassle across platforms and with escaping etc. than they're worth.

also PARSE_TEENY already ignores commented lines:

```make
PARSE_TEENSY = $(shell grep -v "^\#" "$(BOARDS_TXT)" | grep $(1).$(2) | cut -d = -f 2- )
```

DaWelter:
All right. Didn't see that before. Now removed the `^` again.

As to spaces, I would appreciate if my quotation marks stayed in. I mean they do no harm, do they? And it works for me!

sej7278:
well quoted spaces will break on windows at least, its just not a good idea. [article](https://www.cmcrossroads.com/article/gnu-make-meets-file-names-spaces-them)

DaWelter:
Oh well, okay then they shall be gone.

sej7278:
i've merged now, thanks for the PR (and patience!)

sudar:
@DaWelter Thanks for the PR.

@sej7278 As always thanks for going through the PR and for merging it.

Genoil:
This change auto selects overclocked speeds on Teensy 4 and 4.1, causing instability on (at least) SDIO card operation. Better to select the default speed. The code below is a first attempt to fix this, but this will still select 96Mhz (overclock) for Teensy 3.1/3.2. 

`
ifndef F_CPU
  ifndef BOARD_SUB
    # Obtain sequence of supported frequencies.
    SPEEDS = $(call PARSE_BOARD,"$(BOARD_TAG),menu.speed.*")

    # we're only interested in the default speed (first in the list)
    BOARD_SUB = $(shell grep -Eo '^[0-9]+' <<< "$(SPEEDS)")
  endif
  F_CPU := $(call PARSE_BOARD,$(BOARD_TAG),menu.speed.$(BOARD_SUB).build.fcpu)
endif`

PR #485, opened 2017-03-06
Fixes for Fedora: optional LTO, ccache compatibility

timn:
Currently, Arduino.mk cannot easily be used on Fedora 25.
- There is a bug that makes avr-gcc crash with LTO enabled and using the Serial library (no further investigation as into why, it's just that at this time it is broken)
- When enabling ccache one currently has to override the executables explicitly as the base assumption is that everything is combined in one single avr tools directory

This pull request overcomes these issues by:
- make LTO optimization optional through the USE_LTO variable to override its use
- disable LTO by default if Fedora is detected (all other systems operate as before)
- verify tool paths and if the paths do not resolve (i.e., there is no avr tools dir but the tools are in the system path) try to find it in the PATH
- for a system path avrdude, try to guess the config file to be in the system config directory, too (did not set any before)

As far as I can tell the changes are backward compatible. The LTO fix is crucial, the path fix is for convenience and to have to override executables less often.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -902,6 +902,9 @@ ifndef OVERRIDE_EXECUTABLES
     AR      = $(AVR_TOOLS_PATH)/$(AR_NAME)
     SIZE    = $(AVR_TOOLS_PATH)/$(SIZE_NAME)
     NM      = $(AVR_TOOLS_PATH)/$(NM_NAME)
+
+    # Check if the default exist, otherwise try to find in PATH
+    $(foreach C,CC CXX AS OBJCOPY OBJDUMP AR SIZE NM,$(if $(wildcard $C),,$(eval $C = $$(shell type -p $$($(C)_NAME)))))
 endif
 
 REMOVE  = rm -rf
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1046,7 +1046,12 @@ ifndef AR_NAME
 endif
 
 ifeq ($(shell expr $(CC_VERNUM) '>' 490), 1)
-    USE_LTO ?= 1
+    DISTRO=$(if $(wildcard /etc/os-release),$(shell sed -n "/^ID=/ s/ID=//p" /etc/os-release),unknown)
+    ifeq ($(DISTRO),fedora)
+        USE_LTO ?= 0
+    else
+        USE_LTO ?= 1
+    endif
 endif
 
 
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1039,9 +1039,17 @@ ifndef AR_NAME
     endif
 endif
 
+ifeq ($(shell expr $(CC_VERNUM) '>' 490), 1)
+    USE_LTO ?= 1
+endif
+
+
 ifndef CFLAGS_STD
     ifeq ($(shell expr $(CC_VERNUM) '>' 490), 1)
-        CFLAGS_STD      = -std=gnu11 -flto -fno-fat-lto-objects
+        CFLAGS_STD      = -std=gnu11
+        ifeq ($(USE_LTO),1)
+            CFLAGS_STD += -flto -fno-fat-lto-objects
+        endif
     else
         CFLAGS_STD        =
     endif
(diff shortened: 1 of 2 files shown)


sej7278:
i'm not so keen on a bunch of changes for one linux distro version, is there not a bug report about LTO crashing avr-gcc? is fedora still doing all that weird stuff with ccache symlinks?

timn:
I have not yet filed a bug on Fedora, didn't have the time. Will do and then link. However, in the meantime, it would make sense to still provide a working system.

The "weird" ccache stuff is actually exactly what ccache asks you to do (cf. https://ccache.samba.org/manual.html#_run_modes). The only difference to, say, Ubuntu, is that once installed it enables it for all users.

I would argue that one task for a build system is to mitigate system differences to keep them away form the developer (having written the build system for our robot software [Fawkes](https://www.fawkesrobotics.org) this is what we tried). For arduino-mk there are already many assumptions being made, my patches make some others (compatible ones) to suit other systems. LTO is one such assumption. Just providing a possibility to turn it off actually softens that. It's an optional flag with this. See the fedora-specific default as preliminary until the issue has been resolved, after which it's just an option flag.

And yes, I'll fix the build fails. I think it's not using bash as shell for make and this can be fixed not using "type -d" but "which".

Arduino.mk
```diff
@@ -444,6 +444,9 @@ ifndef AVR_TOOLS_DIR
         ifdef SYSTEMPATH_AVR_TOOLS_DIR
             AVR_TOOLS_DIR = $(SYSTEMPATH_AVR_TOOLS_DIR)
             $(call show_config_variable,AVR_TOOLS_DIR,[AUTODETECTED],(found in $$PATH))
+            ifndef AVRDUDE_CONF
+                AVRDUDED_CONF := $(call dir_if_exists,/etc/avrdude/avrdude.conf)
```
sej7278: AVRDUDE**D** typo

is /usr/share/arduino/hardware/tools/avrdude.conf not just a symlink to /etc/avrdude.conf on fedora (like debian)? and is it really /etc/avrdude/avrude.conf or do you mean /etc/avrdude.conf ?

Arduino.mk
```diff
@@ -1039,9 +1045,22 @@ ifndef AR_NAME
     endif
 endif
 
+ifeq ($(shell expr $(CC_VERNUM) '>' 490), 1)
```
sej7278: seems like the fedora bug should just be documented and users told to set USE_LTO=0 rather than adding code to detect fedora that at some point will no longer be needed

sej7278:
Probably worth updating the docs regarding working around ccache on fedora:

https://github.com/sudar/Arduino-Makefile/tree/master/packaging/fedora

Arduino.mk
```diff
@@ -899,6 +902,9 @@ ifndef OVERRIDE_EXECUTABLES
     AR      = $(AVR_TOOLS_PATH)/$(AR_NAME)
     SIZE    = $(AVR_TOOLS_PATH)/$(SIZE_NAME)
     NM      = $(AVR_TOOLS_PATH)/$(NM_NAME)
+
+    # Check if the default exist, otherwise try to find in PATH
+    $(foreach C,CC CXX AS OBJCOPY OBJDUMP AR SIZE NM,$(if $(wildcard $C),,$(eval $C = $$(shell type -p $$($(C)_NAME)))))
```
sej7278: does this actually work (when using which) it doesn't seem to populate the variables when i try, so you end up with: 

```
/bin/sh: 1: CC_NAME: not found
/bin/sh: 1: CXX_NAME: not found
/bin/sh: 1: AS_NAME: not found
/bin/sh: 1: OBJCOPY_NAME: not found
/bin/sh: 1: OBJDUMP_NAME: not found
/bin/sh: 1: AR_NAME: not found
/bin/sh: 1: SIZE_NAME: not found
/bin/sh: 1: NM_NAME: not found
```

sej7278:
p.s. on debian they do the ccache symlink thing but don't enable it by default like they seem to on fedora, so you have to set /usr/lib/ccache at the beginning of your $PATH (yuk!)

also /bin/sh is the default shell in make, on debian that just happens to be symlinked to dash, i expect on fedora its bash (it is on centos).

sej7278:
not sure we need this anymore - lto serial issue is fixed with recent commit. is ccache still an issue?

nspring:
I'd like to see some way to disable LTO.  This LTO feature seems to prevent including the mmcu section that configures simavr, (with LDFLAGS including, for example, -Wl,--undefined=_mmcu,--section-start=.mmcu=0x910000), at least when including a .c file that includes only the mmcu definitions.  

The USE_LTO flag seems like a good enough solution to me.

sudar:
@timn 

Thanks for the PR.

I am closing this PR without merging based on @sej7278 comment about LTO issue being fixed in Fedora.

Feel free to re-open if you have additional information or details for this PR.

PR #496, opened 2017-04-28
Add support for nonstandard makefile on upload

jbott:
Allows for per-device makefiles

For example, Makefile-uno and Makefile-mega. This correctly spawns the
submake command with the appropriate makefile specified.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1451,12 +1451,12 @@ error_on_caterina:
 # Use submake so we can guarantee the reset happens
 # before the upload, even with make -j
 upload:		$(TARGET_HEX) verify_size
-		$(MAKE) reset
-		$(MAKE) do_upload
+		$(MAKE) -f $(firstword $(MAKEFILE_LIST)) reset
+		$(MAKE) -f $(firstword $(MAKEFILE_LIST)) do_upload
 
 raw_upload:	$(TARGET_HEX) verify_size
-		$(MAKE) error_on_caterina
-		$(MAKE) do_upload
+		$(MAKE) -f $(firstword $(MAKEFILE_LIST)) error_on_caterina
+		$(MAKE) -f $(firstword $(MAKEFILE_LIST)) do_upload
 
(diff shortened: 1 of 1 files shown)


sudar:
@sej7278 

Interested to know your thoughts on this PR.

sej7278:
not sure why you'd not have a different makefile for all recipes in that case - or just use a different BOARD_TAG, but i guess it won't hurt.

i can't really see a use case where you'd just upload to different boards (when you'd need to compile for each board already anyway).

jbott:
I found the project I used this in - It's been a year so sadly I don't remember the context this came up in.

**Makefile**
```
BOARD_TAG    = uno
CPPFLAGS    += -DPING_MASTER
MONITOR_PORT = /dev/tty.usbmodem1421

include ./common.mk
include ~/Dev/Arduino-Makefile/Arduino.mk
```

**Makefile-mega**
```
BOARD_TAG    = mega
BOARD_SUB    = atmega2560
MONITOR_PORT = /dev/tty.usbmodem141

include ./common.mk
include ~/Dev/Arduino-Makefile/Arduino.mk
```

**common.mk**
```
ARDUINO_LIBS =
MONITOR_CMD  = cu
```

I'm essentially using each makefile as a build configuration management system - one was the master and the other the client. Looking back on this code a year later, I don't think this is a great way to accomplish my goal.

Regardless, the symptom here is that `make -f Makefile-mega`, would cause the sub-make call to invoke `make reset` and reset via the serial port referenced in the default makefile (or whatever related command was invoked). This patch changes that behavior so it will always invoke make via `make -f <root makefile path>`.

I'm neutral on whether this is a required change. It seems to make more sense than the default behavior, however I'm not convinced that the additional complexity should be introduced into this repo.

sudar:
@jbott 

I am closing this for now, since even I am not convinced if this is the best approach.

Feel free to reopen it if you find a better way to solve it or find a really useful usecase.

Issue #499, opened 2017-05-28
Add support of externally supported boards like Pro Micro


MilanV:
Some companies provide an official support of their boards, like [Sparkfun does](https://github.com/sparkfun/Arduino_Boards/). Arduino IDE knows to add that as an "addon". How can I add this to the arduino-mk as well, please? Either manually or, better, a script taking that .json file and doing proper actions so, as a result, 'arduino-mk show-boards' shows those newly defined ones among the original ones.

sej7278:
See [Blink example](https://github.com/sudar/Arduino-Makefile/blob/master/examples/Blink/Makefile#L41)

You simply download the board's core to your hardware directory instead of using the Boards Manager method. So you end up with something like this in ~/sketchbook/hardware/

```
sparkfun/
└── avr
    ├── boards.txt
    ├── bootloaders
    ├── driver
    ├── platform.txt
    ├── signed_driver
    └── variants
```

And then your Makefile should look something like this:

```make
MONITOR_PORT        = /dev/ttyACM0
BOARD_TAG           = promicro16
ALTERNATE_CORE      = sparkfun

include /usr/share/arduino/Arduino.mk
```

Note with the Pro Micro you can just use ```BOARD_TAG=leonardo``` the sparkfun core doesn't add anything much.

MilanV:
Thank you for your fast and detailed answer.

I start with your last note: I used "leonardo" board definition at the beginning, and the sketch was uploaded and worked. But the ttyACM device disappeared and, since then, the only moment to see it was those 3 seconds when bootloader creates it. After that, not even IDE could help (its detection of the connected board is too slow). So it would be more than useful to have Arduino-Makefile working and this case documented in this package so next arduino-greenhorns don't fall into that deep water as I did :) It's a good message everything is already implemented and it's just a matter of the documentation.

Here is what I found when trying your advice:

* the current "package" from sparkfun contains avr/<version>/... drectory tree, I had to remove that <version> subdirectory level (move everything one dir up) because Arduino.mk was confused by that

* there is no "promicro16" or "promicro8" model (any longer), they are defined as "cpu menu" of one "promicro" model. To use that with Arduino.mk, we need this configuration (8MHz variant):
    ALTERNATE_CORE      = sparkfun
    BOARD_TAG           = promicro
    BOARD_SUB           = 8MHzatmega32U4

[...]

PR #501, opened 2017-06-01
Document how to setup Makefile for 3rd party boards

MilanV:
Add a documentation for the others not to fall into the #499 issue

### DIFF
--- a/examples/MakefileExample/Makefile-3rd_party-board.mk
+++ b/examples/MakefileExample/Makefile-3rd_party-board.mk
@@ -31,7 +31,7 @@
 ### 3. Create this Makefile (use your vendor/package name)
 
 ###        ALTERNATE_CORE      = sparkfun
-###        include ~/Arduino-Makefile/Arduino.mk
+###        include $(HOME)/Arduino-Makefile/Arduino.mk
 
 ### 4. run 'make show_boards'
 ###    check that you can see (only) boards provided by this vendor
@@ -52,5 +52,5 @@ ARDUINO_SKETCHBOOK  = $(HOME)/sketchbook
 ALTERNATE_CORE      = sparkfun
 BOARD_TAG           = promicro
 BOARD_SUB	    = 8MHzatmega32U4
-include $(ARDMK_DIR)/Arduino.mk
+include /usr/share/arduino/Arduino.mk
 
--- a/examples/MakefileExample/Makefile-3rd_party-board.mk
+++ b/examples/MakefileExample/Makefile-3rd_party-board.mk
@@ -1,0 +1,56 @@
+### DISCLAIMER
+### This is an example Makefile and it MUST be configured to suit your needs.
+### For detailed explanations about all of the available options, please refer
+### to https://github.com/sudar/Arduino-Makefile/blob/master/arduino-mk-vars.md
+
+### How to setup a project using a board definition provided by the 3rd party
+### =========================================================================
+
+### Some vendors provide definitions/configuration of their boards separately,
+### as so-called addon packages. Originally, they are supposed to be used in
+### the Arduino IDE but they can be used with Arduino-Makefile as well:
+
+### 1. get the package from the vendor
+###    if they provide .json file, look into it and take the URL of the package
+###       archive from there
+###
+### 2. extract the package into your ARDUINO_SKETCHBOOK directory
+###    you have to end with the directory structure like this, in your
+###    ARDUINO_SKETCHBOOK directory (sparkfun is the vendor name):
+
+###        hardware/
+###        ├── sparkfun/
+###        │   └── avr/
+###        │       ├── boards.txt
+###        │       ├── bootloaders/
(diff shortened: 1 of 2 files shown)


sudar:
@sej7278 

What are your thoughts on this PR?

examples/MakefileExample/Makefile-3rd_party-board.mk
```diff
@@ -0,0 +1,56 @@
...
+###        │       ├── bootloaders/
+###        │       ├── driver/
+###        │       ├── platform.txt
+###        │       ├── signed_driver/
+###        │       └── variants/
+
+### 3. Create this Makefile (use your vendor/package name)
+
+###        ALTERNATE_CORE      = sparkfun
+###        include ~/Arduino-Makefile/Arduino.mk
```
sej7278: don't use tilde

examples/MakefileExample/Makefile-3rd_party-board.mk
```diff
@@ -0,0 +1,56 @@
...
+#####
+
+### The basic configuration should be done now.
+### The example follows:
+
+ARDUINO_SKETCHBOOK  = $(HOME)/sketchbook
+ALTERNATE_CORE      = sparkfun
+BOARD_TAG           = promicro
+BOARD_SUB	    = 8MHzatmega32U4
+include $(ARDMK_DIR)/Arduino.mk
```
sej7278: ARDMK_DIR isn't defined, can we use an absolute path e.g /usr/share/arduino/Arduino.mk ?

MilanV:
Thanks for comments, I fixed it.

sej7278:
Looks good to me now thanks! :+1: 

People aren't looking at https://github.com/sudar/Arduino-Makefile/blob/master/examples/Blink/Makefile i guess, as that has various board examples too.....

MilanV:
I still see a room for improvement here. I believe a new document as a root of the documentation would be helpful. The one describing the whole ecosystem: terminology, list of config files and which does what, what is used from the Arduino IDE etc. Then, the document I created would become just one chapter ("Using a board not defined in he standard Arduino package"). For example, I found a description of Makefile targets only thanks to a TAB completion on Linux command-line, which suggested "help" as one of possible targets of this Makefile :)

But that is definitely out of my knowledge, I'm still an arduino greenhorn :) Just spreading the idea...

sej7278:
Yes, we could probably make a whole .MD file about 3rd party boards, merging now thanks

zvipesh:
WINDOWS-7:

Is it possible to list the required PATH setting for the Arduino IDE 1.8.2 and any other tool required in order to generate the "bootloaderxxx.hex" ?

Do we need to install additional tools besides the IDE installed files ?

I keep getting error indicating the MAKE function does not finds the destination file .
something like the following will be of great help.

for example:
1. Install IDE 1.8.2 in c:\program files (x86)
2. Set PATH as follows:
   a. in control panel - Control Panel\System and Security\System  click the "Advanced system setting"
   b. Click the tab : Environment variables"  
   c. In the "user variables for xx" select  PATH and click EDIT tab
   d.  Add   to PATH:  C:\Program Files (x86)\Arduino\hardware\tools\avr\bin;
         : 
         : < list all PATHs needed to run the compiler etc >
   f.  In the "system variables" select  PATH and click EDIT tab
   g.  Add   to PATH:   < List PATH parameters >

   h. when done - press OK
   i.  Press OK again in the System Properties.

my "user variables"  PATH parameters are:
C:\ProgramData\Oracle\Java\javapath;C:\Program Files (x86)\GnuWin32\bin;C:\Program Files (x86)\AVRDUDE 6.3;C:\Program Files (x86)\Arduino;c:\Program Files (x86)\Arduino\hardware\arduino\avr\bootloaders\stk500v2;

and the "system variables are:
C:\Program Files (x86);C:\Program Files (x86)\GnuWin32\bin;C:\Program Files (x86)\Arduino\hardware\tools\avr\bin;C:\WinAVR-20100110\utils\bin;C:\Program Files (x86)\AVRDUDE 6.3;

It all worked  until I had a window update yesterday , tried several iterations to fix the PATH  ..
no luck ..

this is why I suggest there should be a "specific guide" since the MAKE function is not working
when trying for example ( in installed  IDE 1.8.2  ) when in the c:\...\arduino\hardware\arduino\avr\bootloaders\stk500v2 folder and executing MAKE
it does not works. < tried it after a new fresh fresh installation of the IDE >

make_bootloader generation 6-20-2017

sej7278:
We don't really support windows (it does work apparently but we don't really have a way to test it) but would appreciate a PR if someone fancies documenting how to do it.

zvipesh:
what do you support ?  is it lynix ? OS-x ?

sej7278:
linux and osx are the platforms that most of the contributors to this project use, windows apparently works, as do the bsd's. i suggest you search the issues as there are other windows users who have commented.

PR #502, opened 2017-06-04
Support for generation of project tags file

tuna-f1sh:
Considering the number of project files spread in different locations when developing an Arduino project, proper use of tags can be difficult; resolving beyond local functions.

I've added automatic generation of a tags file, which includes:

* Standard source in project dir (.c, .cpp, .h)
* Arduino source in project dir (.ide, .pde)
* Arduino core based on detected project core from Arduino install.
* Included Arduino libraries from user library folder.

As a Vim user I find this hugely useful and think it would be a useful addition for others. Target has been added as `make tags`. There may be some better practices, for some things (I'm not sure about using `pwd` in the shells commands?). I've included a few variables, see updates variables file.

John.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1320,7 +1320,7 @@ ifndef TAGS_FILE
     TAGS_FILE = tags
 endif
 
-# ctags command: append file with user options before
+# ctags command: append, flags unsort (as will be sorted after) and specify filename
 CTAGS_CMD = $(CTAGS_EXEC) $(CTAGS_OPTS) -auf
 
 ########################################################################
@@ -1580,7 +1580,9 @@ generated_assembly: generate_assembly
 
 .PHONY: tags
 tags:
-	rm -f $(shell pwd)/$(TAGS_FILE)
+ifneq ($(words $(wildcard $(TAGS_FILE))), 0)
+	rm -f $(TAGS_FILE)
+endif
 	@$(ECHO) "Generating tags for local sources (INO an PDE files as C++): "
 	$(CTAGS_CMD) $(TAGS_FILE) --langmap=c++:.ino --langmap=c++:.pde $(LOCAL_SRCS)
 ifneq ($(words $(ARDUINO_LIBS)), 0)
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1308,6 +1308,22 @@ $(OBJDIR)/%.sym: $(OBJDIR)/%.elf $(COMMON_DEPS)
 	$(NM) --size-sort --demangle --reverse-sort --line-numbers $< > $@
 
 ########################################################################
+# Ctags
+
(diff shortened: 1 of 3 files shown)


Arduino.mk
```diff
@@ -1562,6 +1578,21 @@ generate_assembly: $(OBJDIR)/$(TARGET).s
 generated_assembly: generate_assembly
 		@$(ECHO) "\"generated_assembly\" target is deprecated. Use \"generate_assembly\" target instead\n\n"
 
+.PHONY: tags
+tags:
+	rm -f $(shell pwd)/$(TAGS_FILE)
+	@$(ECHO) "Generating tags for local sources (IDO an PDE files as C++): "
```
sej7278: should this be:

I**N**O an**d** PDE files as C++

tuna-f1sh:
Thanks good spot, now fixed.

Arduino.mk
```diff
@@ -1562,6 +1578,21 @@ generate_assembly: $(OBJDIR)/$(TARGET).s
 generated_assembly: generate_assembly
 		@$(ECHO) "\"generated_assembly\" target is deprecated. Use \"generate_assembly\" target instead\n\n"
 
+.PHONY: tags
+tags:
+	rm -f $(shell pwd)/$(TAGS_FILE)
```
sej7278: i'm a bit worried about this accidentally deleting stuff if $(TAGS_FILE) is undefined, is it just a single file?

if so then ```rm -f $(TAGS_FILE)``` i guess is a bit safer.....?

also is this going to be in the same directory as the Makefile and .ino, or in $(OBJDIR)? i.e. will "make clean" remove it?

sej7278:
@sudar is more of a vim guy, care to chime in?

Arduino.mk
sudar: I agree. We need a guard condition to prevent accidental deletion.

tuna-f1sh:
Thanks. I've taken this into account - I agree that even with the definition check of the tags file, it could lead to problems. I use your syntax plugin @sudar so I thought you might like it 👍 .

It's now `rm -f $(TAGS_FILE)` and only if the TAGS_FILE exists. One could get away without deleting the tags file completely as ctags replaces by default but I make use of the append flag to do a multi-line operation. It is common practice for the tags file to be in the root of the project directory, so `make clean` will not remove it - I think this is desirable however.

sej7278:
odd, i'm sure merging #524 conflicts with this, but github isn't picking it up.

does it only pick up changes from master when a PR is updated?

tuna-f1sh:
I've merged the upstream to check but there were no conflicts.

sej7278:
looks good to me 👍

sej7278:
can we merge this now @sudar ?

sudar:
@sej7278 Just merged it.

Issue #497, opened 2017-05-28
Not detecting library directory


chrisspen:
I have a number of extra libraries installed in `/home/$USER/Arduino/libraries`, which I can compile with from inside the Arduino IDE. However, if I try to compile with the makefile, I get the error:

    The following libraries specified in ARDUINO_LIBS could not be found (searched USER_LIB_PATH and ARDUINO_LIB_PATH)

Please add support for custom library paths.

I tried setting USER_LIB_PATH, but this doesn't appear to support `$USER` or `~`, and my application doesn't allow hard-coding the username in the path, since it's not a static value.

sej7278:
Use $HOME instead of ~ in Makefiles

chrisspen:
Ok. In that case, the docs should probably be updated. [This](https://github.com/sudar/Arduino-Makefile/blob/master/arduino-mk-vars.md#user_lib_path) explicitly says `~` should work. But you're right, it looks like `$(HOME)` and `$(USER)` variables are available.

Is it possible to include relative directories? I have a library in the same directory as my makefile, and to simplify installation and maintainability, I'd prefer to avoid having to copy it into my sketchbook.

However, if I used the setting:

    USER_LIB_PATH = .

it still complains that it can't find the library.

sej7278:
have a look through the other issues, there's a few regarding libraries in the sketch directory, don't think its possible

chrisspen:
Is appending multiple paths to `USER_LIB_PATH` supported? [This blog post](http://hackaday.com/2015/10/01/arduino-development-theres-a-makefile-for-that/) raves about Arduino-Makefile, claiming it has this ability, but when I use their syntax, e.g.

    USER_LIB_PATH = /home/$(USER)/Arduino/libraries
    USER_LIB_PATH += .

it causes make to not find any libraries.

sej7278:
fixed the tilde documentation issue in PR #500 now merged

chrisspen:
My end goal is to allow inclusion of a local `./lib` directory, so I don't have to copy every project's libraries into the global sketchbook directory, since that becomes difficult to maintain and inevitably results in conflicts.

Is there any way to currently accomplish this with Arduino-Makefile?

ladislas:
@chrisspen - I did something similar with the [bare arduino project](https://github.com/ladislas/Bare-Arduino-Project)

ladislas:
[...]

PR #503, opened 2017-06-19
Added ARDUINO_SKETCHBOOK to include list to improve search for items in ARDUINO_LIBS

chrisspen:
This fixes the problems discussed in issue #497.

In short, there's currently no reliable way to compile code with Arduino-Makefile that uses an include like:

    #include "Adafruit_BNO055/Adafruit_Sensor.h"

If the library `Adafruit_BNO055` is located `USER_LIB_PATH`, then Arduino-Makefile auto-generates the include `-I$(USER_LIB_PATH)/Adafruit_BNO055`, which allows you to make includes like:

    #include "Adafruit_Sensor.h"

but this could lead to conflicts when other libraries use the same filename.

You can get it to compile for a single library by setting `USER_LIB_PATH = /full/path/to/Adafruit_BNO055/..` but since `USER_LIB_PATH` is used as the base for constructing all other include search paths, it'll cause an error for all other libraries in your `ARDUINO_LIBS`.

The other advantage of this modification is that it makes custom `.lib/` directories trivially easy to implement by simply settings this value in your `ARDUINO_SKETCHBOOK`. Using a custom sketchbook directory per project allows better encapsulation and prevents maintenance headaches and version conflicts. Howeever, by not using the sketchbook directory as an include path itself, you greatly limit its usefulness.

Before using Arduino-Makefile, I used the now defunct [Inotool](https://github.com/amperka/ino), which assumes this relative path in your project by default. The lack of this feature in Arduino-Makefile is the only thing stopping me from adopting it for all my Arduino projects.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1014,7 +1014,7 @@ endif
 
 # Using += instead of =, so that CPPFLAGS can be set per sketch level
 CPPFLAGS      += -$(MCU_FLAG_NAME)=$(MCU) -DF_CPU=$(F_CPU) -DARDUINO=$(ARDUINO_VERSION) $(ARDUINO_ARCH_FLAG) -D__PROG_TYPES_COMPAT__ \
-        -I$(ARDUINO_CORE_PATH) -I$(ARDUINO_VAR_PATH)/$(VARIANT) \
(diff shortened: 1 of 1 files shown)


sudar:
@chrisspen 

Thanks for the PR. Currently the tests are failing and your branch also has merge conflict.

Can you take a look at it?

PR #516, opened 2017-08-08
Quote the prefix tag in the space_pad_to function

szeder:
POSIX shells treat an unquoted [abcd] as a pattern bracket expression
and apply it for filename expansion.  This kicks in and causes
troubles in the space_pad_to function when it's called to pad prefix
tags like '[AUTODETECT]', because the argument holding such a tag is
passed to a shell unquoted.  The result is funny output when the
directory containing the Makefile also contains any files or
directories, whose name is a single upper-case character that can be
found in any prefix tags:

```
  $ touch A B C D
  $ make
  -------------------------
  Arduino.mk Configuration:
  - A C D                CURRENT_OS = LINUX
  - [USER]               ARDUINO_DIR = /home/szeder/src/arduino/arduino-1.8.1
  - [USER]               ARDMK_DIR = /home/szeder/src/arduino/Arduino-Makefile
  - A C D                ARDUINO_VERSION = 181
  - A D                  ARCHITECTURE = avr
  - A D                  ARDMK_VENDOR = arduino
  - A D                  ARDUINO_SKETCHBOOK =
  <snip>
```

Prevent this by quoting space_pad_to's parameter when passing it to a
shell.

### DIFF
--- a/Common.mk
+++ b/Common.mk
@@ -14,7 +14,7 @@ runscript_if_exists =                                                          \
 
 # For message printing: pad the right side of the first argument with spaces to
 # the number of bytes indicated by the second argument.
-space_pad_to = $(shell echo $(1) "                                                      " | head -c$(2))
+space_pad_to = $(shell echo "$(1)                                                       " | head -c$(2))
 
 # Call with some text, and a prefix tag if desired (like [AUTODETECTED]),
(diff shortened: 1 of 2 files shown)


sej7278:
👍 looks good to me thanks

PR #522, opened 2017-08-28
Search for headers and sources in all libraries' subfolders

### LINKED ISSUE #521: Search for headers and source files in subdirectories other than utility
Currently, [`get_library_includes`](https://github.com/sudar/Arduino-Makefile/blob/master/Arduino.mk#L927) and [`get_library_files`](https://github.com/sudar/Arduino-Makefile/blob/master/Arduino.mk#L934) are only searching the the `utility` subfolder.

But it sometimes happen that directories have different names like [ChRt](https://github.com/greiman/ChRt/tree/master/src).

Adding the following lines fixes everything! 👍  And we could even remove the `utility` part as the wildcard takes care of that.

```diff
diff --git a/Arduino.mk b/Arduino.mk
index 50d60ee..78d1eac 100644
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -926,14 +926,15 @@ rwildcard=$(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2) $(filter $(subst
 # Gets include flags for library
 get_library_includes = $(if $(and $(wildcard $(1)/src), $(wildcard $(1)/library.properties)), \
                            -I$(1)/src, \
-                           $(addprefix -I,$(1) $(wildcard $(1)/utility)))
+                           $(addprefix -I,$(1) $(wildcard $(1)/utility)) \
+                           $(addprefix -I,$(1) $(dir $(wildcard $(1)/*/))))

 # Gets all sources with given extension (param2) for library (path = param1)
 # for old (1.0.x) layout looks in . and "utility" directories
 # for new (1.5.x) layout looks in src and recursively its subdirectories
 get_library_files  = $(if $(and $(wildcard $(1)/src), $(wildcard $(1)/library.properties)), \
                         $(call rwildcard,$(1)/src/,*.$(2)), \
-                        $(wildcard $(1)/*.$(2) $(1)/utility/*.$(2)))
+                        $(wildcard $(1)/*.$(2) $(1)/*/*.$(2) $(1)/utility/*.$(2)))

 # General arguments
 USER_LIBS      := $(sort $(wildcard $(patsubst %,$(USER_LIB_PATH)/%,$(ARDUINO_LIBS))))```

### PR DESCRIPTION
**Pull request by ladislas**
fixes #521 

I've kept the `utility` folder to make it clearer but I can remove it if you want.

let me know.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -924,16 +924,17 @@ rwildcard=$(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2) $(filter $(subst 
 # and library.properties file
 
 # Gets include flags for library
-get_library_includes = $(if $(and $(wildcard $(1)/src), $(wildcard $(1)/library.properties)), \
+get_library_includes = $(sort $(if $(and $(wildcard $(1)/src), $(wildcard $(1)/library.properties)), \
                            -I$(1)/src, \
-                           $(addprefix -I,$(1) $(wildcard $(1)/utility)))
+                           $(addprefix -I,$(1) $(wildcard $(1)/utility)) \
+                           $(addprefix -I,$(1) $(dir $(wildcard $(1)/*/)))))
 
 # Gets all sources with given extension (param2) for library (path = param1)
 # for old (1.0.x) layout looks in . and "utility" directories
 # for new (1.5.x) layout looks in src and recursively its subdirectories
-get_library_files  = $(if $(and $(wildcard $(1)/src), $(wildcard $(1)/library.properties)), \
+get_library_files  = $(sort $(if $(and $(wildcard $(1)/src), $(wildcard $(1)/library.properties)), \
                         $(call rwildcard,$(1)/src/,*.$(2)), \
-                        $(wildcard $(1)/*.$(2) $(1)/utility/*.$(2)))
+                        $(wildcard $(1)/*.$(2) $(1)/utility/*.$(2) $(1)/*/*.$(2))))
 
 # General arguments
 USER_LIBS      := $(sort $(wildcard $(patsubst %,$(USER_LIB_PATH)/%,$(ARDUINO_LIBS))))
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,6 +5,7 @@ The following is the rough list of changes that went into different versions.
 I tried to give credit whenever possible. If I have missed anyone, kindly add it to the list.
 
 ### In Development
+- Tweak: recursively look for .h and .c/.cpp files in library subfolders (issue #521). (https://github.com/ladislas)
 - Fix: Add -fno-devirtualize flag to workaround g++ segfault bug (issue #486). (https://github.com/sej7278)
 - Fix: Quote the prefix tag in the space_pad_to function
 - Tweak: Set ARDMK_VERSION to 1.6 (https://github.com/sej7278)

### CODE REVIEW

**Comment by sej7278** [comment]
doesn't work for me.

it seems to repeatedly include the same library over and over - including trying to compile examples, which then can't find libraries they're using e.g. RF24

**Comment by ladislas** [comment]
> have you tested it without your auto-lib-detection changes?

yep, I used it with vanilla arduino-makefile.

but I usually remove the `example` subfolders or anything that is not sources.

maybe we can try to exclude the examples?

**Comment by ladislas** [comment]
@sej7278 I was able to reproduce your behavior. it seems to include the same library as many times as there are .h files.

**Comment by ladislas** [comment]
adding `$(sort)` will fixe the issue:

```
# Gets include flags for library
get_library_includes = $(sort $(if $(and $(wildcard $(1)/src), $(wildcard $(1)/library.properties)), \
                           -I$(1)/src, \
                           $(addprefix -I,$(1) $(wildcard $(1)/utility)) \
                           $(addprefix -I,$(1) $(dir $(wildcard $(1)/*/)))))
```

But it will also sort the include order. Can this cause any issues?

**Comment by ladislas** [comment]
from what I'm reading, the order is not important as long as it compiles. It can become an issue when headers depend on other headers that need to be included first.

**Comment by ladislas** [comment]
@sej7278 I've just updated the PR and added the tweak to HISTORY.md

let me know if you need anything more before merging.

**Comment by sej7278** [comment]
it still breaks when the libraries include examples (which all should!) in my case the RF24 library includes examples for the raspberry pi, not just arduino code, which is probably not a good idea, but does highlight the problem:

```
libraries/RF24/examples_RPi/gettingstarted.cpp:22:19: fatal error: cstdlib: No such file or directory
 #include <cstdlib>
                   ^
compilation terminated.
```

**Comment by ladislas** [comment]
Hmm I see.

I agree, they should have examples. Most of the time they are `.ino` files so they won't be looked for.

The problem arises as you've experienced when people start packing up simple Arduino libs with examples for other systems...

I don't know how we could differentiate those in pure Makefile.

With the `find` command though, it would be as easy as: (not sure it works on Windows)

```
find . -type f -name "*.cpp" -not -path "*example*"
```

But then I'm afraid we'll discover that we should also exclude the `test` directory, etc.

ladislas:
We have to decide what's best:

- find files when people organize their library without a `utilitiy` folder
- not try to compile things that don't belong to the library.

I don't mind having that feature just for myself as I'll cleanup the libs if needed. And we can keep the issue with the solution if people hit the same situation.

sej7278:
not sure if the utility issue this is trying to fix is going to occur more than the fix is going to break....

all of the RF24 libraries will break, as well pubsubclient and ArduinoJson due to "test" directories and others

ladislas:
removing test and example is super easy. we can add them as they come. 

if it's too complicated I'll keep that in my fork as I know exactly what I'm using :)

sej7278:
i know of:

tests, test, extras, examples, examples_RPi, RPi, Examples, travis, ssl, docs, scripts, linux, boards_json, Adapters, Blynk, fuzzing, third-party.

its not maintainable to list directories *not* to search.

sudar:
@ladislas 

Thanks for the PR.

I agree with @sej7278 comments and have decided to close this PR without merging.

Please do re-open if you have found a different way to solve this.

Issue #523, opened 2017-09-01
User and system flags


ladislas:
Hi there! --

As of now, it seems like setting `CFLAGS_STD` or `CXXFLAGS_STD` removes a lot of important flags.

This can be confusing for users who just might want to use `-std=gnu++14` but keep the other important flags as is.

I think we should have different variables defined:

- `CFLAGS_STD` & `CXXFLAGS_STD` must be only for `-std=xxx` and provide a sane default
- `CFLAGS_CORE` for `-flto -fno-fat-lto-objects -fdiagnostics-color`
- `CXXFLAGS_CORE` for `-fno-threadsafe-statics -flto -fno-devirtualize -fdiagnostics-color`
- `CFLAGS_USER` & `CXXFLAGS_USER` - for the user to add more flags and still keep the core ones

the `*_USER` flags can be kept empty and the `*_CORE` can also be set to empty if the user wants to provide only his own flags.

What do you think?

sej7278:
yes the *_STD variables got abused a bit recently in the rush to fix LTO and g++ issues, they should revert to just setting the standard e.g. gnu++11

ok so CFLAGS_CORE must not be user-modifiable.

if gcc >4.9 is used, then CFLAGS_CORE must be set with the LTO etc. flags.

if gcc <4.9 is used, then CFLAGS_CORE must be empty.

CFLAGS_USER is down to the user to know what they're doing.

the simpler way of doing it is probably to append them to CFLAGS and not bother with CFLAGS_CORE (or CFLAGS_USER as the user can simply append to CFLAGS too):

```diff
diff --git a/Arduino.mk b/Arduino.mk
index 50d60ee..77313c2 100644
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1045,7 +1045,7 @@ endif
 
 ifndef CFLAGS_STD
     ifeq ($(shell expr $(CC_VERNUM) '>' 490), 1)
-        CFLAGS_STD      = -std=gnu11 -flto -fno-fat-lto-objects -fdiagnostics-color
+        CFLAGS_STD      = -std=gnu11
     else
         CFLAGS_STD        =
     endif
@@ -1056,7 +1056,7 @@ endif
 
 ifndef CXXFLAGS_STD
     ifeq ($(shell expr $(CC_VERNUM) '>' 490), 1)
-        CXXFLAGS_STD      = -std=gnu++11 -fno-threadsafe-statics -flto -fno-devirtualize -fdiagnostics-color
[...]

PR #524, opened 2017-09-02
Moved the LTO etc. flags from CXXFLAGS_STD to CXXFLAGS

sej7278:
See issue #523


sudar:
@sej7278 Looks good to me. Thanks for taking care of this.

sej7278:
OK i'll merge it, but think its going to mess up PR #502 so might need rebasing a bit

Issue #519, opened 2017-08-28
(README) Remove claim that paths must be relative when using Cygwin


birgersp:
Hi,

According to the docs, `ARDUINO_DIR` has to be a relative path when using Windows. This is quite unfortunate, in my opinion. Especially when dealing with collaborative projects, and/or developing the project(s) on multiple computers or installations.

- Why does the path have to be relative?
- Is there really **no** workaround?

tuna-f1sh:
Have you tried using non-relative paths? I made a symbolic link on the root of my drive then set these as defines. Works fine for me using cygwin. Not sure if something has been fixed since this advice.

```
export ARDUINO_DIR=C:/Arduino
export ARDMK_DIR=C:/Arduino-Makefile
```

birgersp:
Thanks for the tip, I'll try some more to get it working. Using relative paths to the directories isn't really an "advice", it's stated pretty clear in the README:

>> In Cygwin in Windows this path must be relative, not absolute (e.g. "../../arduino" and not "/c/cygwin/Arduino").

birgersp:
Non-relative paths works, thank you very much @tuna-f1sh !

Then I guess the statements in the README are false

sej7278:
do non-relative paths only work with a symlink?

tuna-f1sh:
Doesn't seem like it, no. One can't use the default Arduino install directory due to the space in "Program Files (x86)" (escape charactors don't work and all my tests it didn't resolve correctly) but all the other defines work for me as direct paths or symbolic links. It's worth noting that the path must be defined with **/** and not Windows **\\**, so it's a sort of hybrid!

sej7278:
@tuna-f1sh any chance you could do a PR to update the docs as you seem to understand what's involved in windows more than myself/sudar?

perhaps including limitations such as spaces and remove references to requiring symlinks and relative paths.

tuna-f1sh:
I've created https://github.com/sudar/Arduino-Makefile/pull/525

Would be good to get your input too @birgersp.

birgersp:
No objections, PR resolves what I asked for ++

Thanks!

PR #525, opened 2017-09-05
Update Windows documentation to allow non-relative paths

tuna-f1sh:
Relates to issue https://github.com/sudar/Arduino-Makefile/issues/519

I've removed the requirement for relative paths from my own and the issues experience and been a bit more verbose on how to setup the Windows defines. 

I also added support for both the default Cygwin Unix style Python installation and a Windows Python installation (as many people including myself opt to use). The difference is that the Windows Python installation, expects the device port as a COMx (as passed by the user), not ttySx as the Arduino-Makefile currently detects Cygwin and passes. It does this by checking the binary path.

Windows is not my workstation of choice but I made these changes having to setup a work machine. It would be good to get it tested by some other Windows users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,8 +127,7 @@ You need to install Cygwin and its packages for Make, Perl and the following Ser
 Assuming you included Python in your Cygwin installation:
 
 1. download PySerial source package from [https://pypi.python.org/pypi/pyserial](https://pypi.python.org/pypi/pyserial)
-2. extract downloaded package running
-```tar xvzf dowloaded_package_name.tar.gz```
+2. extract downloaded package running `tar xvzf dowloaded_package_name.tar.gz`
 3. navigate to extracted package folder
 4. build and install Python module: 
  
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -205,44 +226,52 @@ On Windows (using MSYS and PuTTY), you might want to set the following extra par
...
 
 Instead of:
 
 ```make
-    ARDUINO_DIR=../../../../../Program\ Files\ \(x86\)/Arduino
+    ARDUINO_DIR=C:/Program\ Files\ \(x86\)/Arduino
 ```
 
-Usefull Variables:
+### Usefull Variables
```
sej7278: Useful

tuna-f1sh:
Thanks, I missed that when I did the title formatting. I picked up a couple of other spelling mistakes. After a bit of thought last night, I've added a catch in Common.mk for the correct Windows defines.

I also fixed the ARDUINO_SKETCHBOOK detection for Windows to use USERPROFILE, rather than HOME. This resolves the path as a C:/, rather than cygdrive, which did not work with the Windows build tools. This means that ARDUINO_SKETCHBOOK no longer requires a definiation so I have removed it from my documentation update.

sej7278:
👍 looks good to me, no idea if it works, but you seem to know, it doesn't break linux ;-)

@sudar any feedback?

tuna-f1sh:
I've checked on my Arch install and main macOS machine too.

sudar:
Thanks @tuna-f1sh and @sej7278

PR #529, opened 2017-09-26
Recognize serial monitors with full path in MONITOR_CMD

szeder:
When assembling the command line for the various supported serial
monitors, `MONITOR_CMD` must match the name of one of the supported
commands to be recognized.  Serial monitors given with leading path
components are not recognized, and a command like

  `make MONITOR_CMD=~/src/picocom/picocom monitor`

errors out as the fallback monitor command is executed instead of the
`picocom`-specific one.  However, sometimes it's necessary to specify a
supported serial monitor with its full path, because e.g. the user
wants to tests a freshly compiled version before installing it.  Sure,
the user could just run the serial monitor directly, but that's
cumbersome because he has to pay attention to use the right baud rate
and USB port.

So strip all leading path components, if present, from `MONITOR_CMD`
using the `nondir` make function before checking whether it's one of
the supported serial monitors.  This way commands like the above would
just work.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1566,15 +1566,15 @@ show_submenu:
 	@$(CAT) $(BOARDS_TXT) | grep -E '[a-zA-Z0-9_\-]+.menu.(cpu|chip).[a-zA-Z0-9_\-]+=' | sort -uf | sed 's/.menu.\(cpu\|chip\)./:/' | sed 's/=/:/' | column -s: -t
 
 monitor:
-ifeq ($(MONITOR_CMD), 'putty')
+ifeq ($(notdir $(MONITOR_CMD)), putty)
 	ifneq ($(strip $(MONITOR_PARAMS)),)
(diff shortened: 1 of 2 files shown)


sej7278:
you sure you're not just hitting the problem with makefiles not liking tildes?

szeder:
Yes:
```
$ make MONITOR_CMD=/usr/local/bin/picocom monitor
/usr/local/bin/picocom /dev/ttyUSB0 57600
Unexpected non-option arguments: 57600 
/home/szeder/src/arduino/Arduino-Makefile/Arduino.mk:1580: recipe for target 'monitor' failed
make: *** [monitor] Error 1
```

because `ifeq ($(MONITOR_CMD), picocom)` will only ever match `picocom` on its own, but never with absolute path.

sej7278:
ok, looks good to me 👍

sudar:
Thanks @szeder and @sej7278

PR #531, opened 2017-10-06
Set `ARDMK_VENDOR` differently if OS is Arch Linux.

doronbehar:


sej7278:
this sounds like arch need to fix their packaging

doronbehar:
You do have a point, but the reason I think this is legitimate is because Arch Linux's philosophy  as for packaging is to make the packages as least monolithic as possible. That's why [since version `1.6.12` (read from line 14)](https://git.archlinux.org/svntogit/community.git/tree/trunk/arduino.install?h=packages/arduino#n14&id=1634f6036117e7fba2b7163533c3b451fb72e228) Arch's developers decided to divide the files consisting of Arduino's IDE to 3 separate packages exactly so it would be possible for packages like [this (`arduino-mk`)](https://aur.archlinux.org/packages/arduino-mk/) could be installed on a machine without unwanted dependencies.

PR #532, opened 2017-10-06
Imorove target `monitor` and add target `stty_params`.

doronbehar:
Use `less` as the default `MONITOR_CMD`. Change comment explanation about
`# Serial Monitor`.

Add `stty_params` target.

Make `stty_params` run a command according to Arch Linux's WiKi article on
Arduino #stty. Make `monitor` target depend on `stty_params`.


sudar:
@doronbehar 

Thanks for the PR.

Unfortunately I didn't get to it on time and it has resulted in some merge conflicts.

Can you please update your branch and fix the merge conflict?

doronbehar:
Rebased against master.

sej7278:
i'm also not sure what advantage less has over screen?

wingunder:
> i'm also not sure what advantage less has over screen?

I totally agree. 

I have no problem to have `less` and `cat` as alternatives, but screen is currently the default and I vote for it to remain like that.

doronbehar:
Personally, as for setting `less` as default, my opinion is that we should put defaults which are installed by default on almost every system. I can revert that if you really disagree with me..

sudar:
@doronbehar 

Changing the default monitor program will also break it for existing users. So I think it would be better to leave the default monitor program. Also if someone wants to change they can easily override the variable.

If you can remove the changes you made to set `less` as the default program and re-submit the PR, we could revisit it again.

doronbehar:
Rebased again against master.

@sudar and @sej7278, I don't know how it goes on your machines but I use Arch Linux and when I run `make monitor` I sometimes get a bunch of gibberish characters and sometimes I get no output at all. With the `stty_params` target, I get a clear output just like I programmed.

sej7278:
the arch arduino package is all sorts of broken so that doesn't surprise me, i've never had a problem on debian/fedora/osx. what's in your ~/.screenrc?

i don't like that stty_params is going to be called in all cases when running "make monitor" even for users without the problem, which could potentially introduce problems for them.

it seems this is changing arduino-mk to work around some sort of archlinux bug.

doronbehar:
Thanks for your comment @sej7278, I might ask the Arch Linux community about the subject and inspect the way the arduino packages are built on other distributions.

I must say though, that I hardly believe there is an automated process that an OS can be configured to set these parameters by default whenever a new Arduino device is connected (perhaps using `udev`?), since there is an `stty` parameter defining the frequency of the communication between the two devices which is obviously different for every project..

Thinking about the frequency parameter, as a 1st idea, I was wondering, perhaps there is a default value being set by your OS's configuration that is identical to the one you are using in **all**  of your Arduino projects? And that is why you never need to run any `stty` command on your machines?

Please leave this PR open for a while, I'll be glad to update you when you or my self will learn something new..

doronbehar:
I have made further tests and it seems the `stty` command is needed only when using `cat`, `tail` or `less` and not `screen`. I've found many suggestions around the internet for invoking a `stty` command:
- https://arduino.stackexchange.com/questions/19002/use-unix-terminal-instead-of-the-monitor-on-arduino-ide
- https://playground.arduino.cc/Interfacing/LinuxTTY
- https://arduino.stackexchange.com/questions/16776/how-to-connect-arduino-device-to-linux
- https://stackoverflow.com/questions/3918032/bash-serial-i-o-and-arduino

I think making the target `stty_params` a dependency of the target `monitor` should improve our safety anyway since printing unprintable characters in a terminal is [known to be dangerous](https://security.stackexchange.com/questions/56307/can-cat-ing-a-file-be-a-potential-security-risk). I'm willing to add more commits that will make the `monitor` target run the `stty_params` target only when a `MONITOR_CMD` other then `screen` is used but I don't see any advantage in this.

Issue #417, opened 2016-03-02
Fuse set problem


thalesmaoa:
Hi, I have all my make configured, but when I try to set fuses, they don't work.

> ISP_LOCK_FUSE_PRE  = 0xfc
> ISP_LOCK_FUSE_POST = 0xfc
> ISP_HIGH_FUSE      = 0xd9
> ISP_LOW_FUSE       = 0xff
> ISP_EXT_FUSE       = 0xfc

But I can only get them working with:

> AVRDUDE_OPTS      = -U hfuse:w:0xD9:m -U lfuse:w:0xFF:m -U efuse:w:0xFC:m -v -v

Besides, AVRDUDE_OPTS run before the erase process, but the fuses are supposed to write after.

Any idea?

sudar:
Can you post your entire makefile and the complete output?

sej7278:
yes can't really tell without the Makefile and output.

i've never tried using lock bits, but your AVRDUDE_OPTS has skipped them too, have you tried removing them from your Makefile to see if the h/l/e fuses get set then?

also avrdude ignores unset bits, so 0xFF should be 0x07, i've come across that problem myself.

ref. http://www.avrfreaks.net/comment/673717#comment-673717

thalesmaoa:
Thanks

Makefile
http://pastebin.com/CShtKUKY

With that file I have the following output:

http://pastebin.com/2J4LXEXf
No fuses written.
If I change:
`AVRDUDE_OPTS 	    = -U hfuse:w:0xD9:m -U lfuse:w:0xFF:m -U efuse:w:0xFC:m -v -v`

I get the following:

http://pastebin.com/yrFwGJxC

Fuses are written but out of order

sej7278:
ah i see the problem - ```make ispload``` doesn't set the fuses, you do that with ```make set_fuses```.

```make burn_bootloader``` uploads the sketch, bootloader and sets the fuses, but we separated out the fuse and ispload recipes.

in your second test case, this is a known [avrdude bug](http://savannah.nongnu.org/bugs/?40794), you can see it reads the 3 fuses in the correct order, but then the last line is out of order - just ignore the last line or upgrade to avrdude 6.1 or later.

```
avrdude: safemode: lfuse reads as FF 
avrdude: safemode: hfuse reads as D9  
avrdude: safemode: efuse reads as FC  
avrdude: safemode: Fuses OK (H:FC, E:D9, L:FF)    <-- only this line is wrong
```

[...]

PR #540, opened 2017-12-22
Moved -e flag for avrdude from set_fuses to ispload

sej7278:
You wouldn't want a chip erase operation when only setting the fuses.

ispload probably should erase the flash+eeprom+lockbits as that's when you upload the sketch.

setting fuses should be just that, not deleting anything.

May help with issue #417 also

My use case is that I want to ispload a sketch, then set the hfuse to 0x5f which will prevent me from being able to ispload afterwards (it uses the reset pin as d5). This allows me to run set_fuses without nuking the script I uploaded with ispload.


sudar:
@sej7278 

Thanks.

PR #530, opened 2017-10-03
ardmk-init - Makefile initialisation and template project utility

tuna-f1sh:
Another pull request - hopefully you're after new features and I'm not just making more work. Let me know if otherwise.

I've created another Python script in bin/, which can create a Makefile for use with Arduino.mk, using either arguments or user prompts. I also added ability to create a starter .ino file and even a traditional directory tree (a bit like *Bare-Arduino-Project*). I created it as I'm frequency creating new projects or needing to compile a Arduino project using Arduino.mk and became tired of creating a Makefile each time. I imagine others will find it useful for the same reasons. It's great for quickly creating an Arduino-Makefile within a library example for example. 

I updated the various supporting files, here is the README extract:

> ## Makefile Generator and Project Initialisation
> 
> `ardmk-init` within the bin/ folder is a utility Python script to create a
> Arduino-mk Makefile for a project and also has option to create a traditional *tree*
> organization (src, lib, bin). It can be used as with commanline arguments or
> prompted - see examples below (append `$ARDMK_DIR/bin/` to command if not on path):
> 
> * Run prompted within current working directory: `ardmk-init`
> * Create Arduino Uno Makefile (useful within a library example): `ardmk-init -qb uno`
> * Create boilerplate Arduino Uno project in current working directory of same
>   name: `ardmk-init -b uno --quiet --project`
> * Create Arduino-mk nano Makefile in current working directory with template .ino: `ardmk-init -b nano -u atmega328 -qtn my-project`
> * See `ardmk-init --help` for more.

Along the way [I've added support for the Atmel Ice ISP](https://github.com/sudar/Arduino-Makefile/commit/5245606f64320f1a595fa252154f55d48e5eee88) (currently available AVR programmer) - it's now supported by the Arduino IDE so makes sense to add. I also [corrected the ctags line](https://github.com/sudar/Arduino-Makefile/commit/1882a8024e4cec0be568436edde461c70039f994), which was exclusively looking at .ino and .pde files as c++ rather than appending to all filetypes - my mistake. I can create these as separate pull requests if you'd like but they are one line changes.

PR #533, opened 2017-10-08
Set `ARDMK_VENDOR` differently if OS is Arch Linux.

doronbehar:
First of all, sorry for the noise. I [already made a PR](https://github.com/sudar/Arduino-Makefile/pull/531) for this branch but I accidentally closed it and I still want to discuss it and hopefully to see it merged.

Thanks.

PR #543, opened 2018-01-08
Arduino Atmel SAM family Support (ARM architecture)

tuna-f1sh:
This branch adds support for the ARM-based Arduinos, (M0 Pro, Zero, MKR1000, etc.)
specifically the Atmel SAM series (Cortex M0).

It works in a similar manner to the Teensy support by adding an additional Makefile to include, that creates defines before the bulk of the Arduino.mk execution. I've deviated slightly from the Teensy method by suggesting the define of `ARDMK_VENDOR = samd` rather than including a different Makefile. This makes it more inline with the current Arduino.mk workflow and also allows printing of user/calculated defines in the STDOUT header. One can still include Sam.mk like the Teensy method but the print information will not been shown.

The commit history will show the main changes - there are not many to Arduino.mk itself so there should be minimal risk of impact to previous operation. Main features and changes: 

* Added `debug_init` and `debug` targets for using GDB, demo here: https://asciinema.org/a/156056.
* The `ARCHITECTURE` define set to 'arm' when using SAM devices, rather than 'sam'. I don't think it was being used previously but seemed confusing as 'sam' isn't an architecture but a chip series. See https://github.com/tuna-f1sh/Arduino-Makefile/blob/sam/Arduino.mk#L326
* `ARDMK_VENDOR` is used to signify the chip series (sam or samd) and used as flag for ARM stuff in Arduino.mk
* OpenOCD, Bossa and Black Magic Probe support added.
* `ARM_TOOLS_DIR` is a new define and used like `AVR_TOOLS_DIR`. I started using the AVR tools one and it worked but then discovered that the native port on the devices can emulate an AVR MCU so AVR tools are still required for some variants. Having a separate define is better for readability and allows both toolchains to be used in different locations - the support packages are often installed in different directories to the base installation.

I've updated the README, arduino-mk-vars and HISTORY so have a look at those too for new defines etc. The work branches from my other [PR](https://github.com/sudar/Arduino-Makefile/pull/530) should that would need merging in first.

The branch is working and includes an example Blink project. I've also successfully tested with a more complex Arduino project, that includes external libs.

One minor issue I've failed to iron out is why `grep` on macOS and Linux has different results when using regex. I understand one is FreeBSD and one is GNU but cannot understand why they should parse regex differently? At the moment I have a switch for macOS but it only captures a single match, rather than multiple. Basically, the boards.txt extra defines includes non-defines that need filtering out (https://github.com/tuna-f1sh/Arduino-Makefile/blob/sam/Sam.mk#L326):

`echo '-D__SAMD21G18A__ {build.usb} -Dtest' | grep -oE '(\B-D)\w+'`

### macOS Output - only last word, incorrect
-Dtest

### Arch Linux Output - both words, correct
-D__SAMD21G18A__ -Dtest

I got round it using \b rather than \B, but this only captures the first word match, could miss multiple flags and does not work on GNU grep.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@ This is a very simple Makefile which knows how to build Arduino sketches. It def
 - Very robust
 - Highly customizable
 - Supports all official AVR-based Arduino boards
+- Supports official ARM-based Arduino boards using Atmel SAM chip family
+  (Cortex M0) and includes on-device debugging targets.
 - Supports chipKIT
 - Supports Teensy 3.x (via Teensyduino)
 - Works on all three major OS (Mac, Linux, Windows)
@@ -313,6 +315,72 @@ See examples/BlinkOpenCM for example usage.
 
 For large Robotis projects, [libmaple](https://github.com/Rhoban/Maple) may be more appropriate, as the OpenCM IDE uses a very old compiler release.
 
+## Arduino ARM Boards
+
+For Arduino boards using ARM architechure, specifically the Atmel SAM series
+(Arduino M0 [Pro], Zero, MKR1000, Feather M0, etc.), first
+install the board support package from the IDE or other distribution channels.
+
+Define`ARDUINO_PACKAGE_DIR` as the root path containing the ARM support
+package (the manufacturer folder) and the `BOARD_TAG` (see `make show_boards`
+for help) within your project Makefile. Include 'Sam.mk' rather than
+  'Arduino.mk' at the end of your file - see examples/ZeroBlink and
+  examples/MZeroBlink for example usage.
+
+**Note**: The Arduino IDE does not install board support packages to
+the base Arduino installation directory (the directory that will work with AVR
+Makefiles). They are generally installed to a '.arduino15/packages' folder in
+the users home directory. This is the reason for the new `ARDUINO_PACKAGE_DIR`
+define. On Windows, the package directory is often in the user home directory
+so advice is to create a symblic link to avoid slash/space in path problems.
+You can also manually install support packages in your Sketchbook 'hardware'
+folder, then define ARDUINO_PACKAGE_DIR as this path.
+
+If using a SAM board from a manufacturer other than Arduino, one must still
+install the Arduino board support as above (unless using externally defined
+toolchain) and then define the location of the manufacturer board support core
+using the ALTERNATIVE_CORE_PATH define. For example: `ALTERNATE_CORE_PATH =
+$(ARDUINO_SKETCHBOOK)/hardware/sparkfun/samd`
+
+The programing method will auto-detect based on the `BOARD_TAG` settings read
+from boards.txt:
+
+Programming using OpenOCD CMSIS-DAP with the Programming/debug USB is
+currently supported (the same method used by the IDE), including burning
+bootloaders. External CMSIS tools such as Atmel Ice will also work with this
+method. Black Magic Probe (BMP) support is also included using GDB for both
+uploading and debugging.
+
+Native USB programing using Bossa (Zero, MKR1000, Feather style bootloaders)
+and avrdude (M0 bootloaders) is supported. The bootloaders on these devices
+requires a double press of the reset button or open/closing the serial port at
+1200 BAUD. The automatic entry of the bootloader is attempted using
+`ard-reset-arduino` when using the general `make upload` target by polling
+attached devices until the bootloader port re-attaches (same method as the
+IDE). On Windows, the USB enumerates as differnt COM ports for the CDC Serial
+and bootloader and these must be defined. On encountering problems, one can
+manually enter the bootloader then upload using the `make raw_upload` target.
+Note that the `make reset` target will enter the bootloader on these devices;
+there is no way to reset over USB.
+
+If using system installed tools, be aware that `openocd` and `bossa` were
+orginally forked for Arduino support and system distributions may not be up
+to date with merged changes. `bossa` must be version 1.7->. `openocd` should
+work but there may be problems at run time
+[ref](https://github.com/pda/arduino-zero-without-ide). Ideally, use the
+support packaged version or compile and install the Arduino fork.
+
+With the ARM chipset and using a CMSIS-DAP tool, on-device debugging is made available:
+
+* `debug_init` and `debug` targets for on-device debugging using GDB. To use
+  this, one must start the GDB server with `make debug_init &`, followed by
+  connecting to the target with `make debug`. If using a Black Magic Probe,
+  one can just use `make debug`. At the moment, a system wide `arm-none-eabi-gdb` must be
+  installed as the one supplied with the Arduino toolchain
+  does not appear to work. 
+* Example usage: https://asciinema.org/a/Jarz7Pr3gD6mqaZvCACQBzqix
+* See the examples/MZeroBlink Makefile for a commented example.
+
 ## Versioning
 
 The current version of the makefile is `1.6.0`. You can find the full history in the [HISTORY.md](HISTORY.md) file
@@ -332,9 +400,8 @@ Also checkout the [contribution guide](CONTRIBUTING.md) for more details.
 
 If you are looking for ideas to work on, then check out the following TODO items or the [issue tracker](https://github.com/sudar/Arduino-Makefile/issues/).
 
-## Limitations / Know Issues / TODO's
+## Limitations / Known Issues / TODO's
 
-- Doesn't support SAM boards yet.
 - Since it doesn't do any pre processing like Arduino IDE, you have to declare all methods before you use them ([issue #59](https://github.com/sudar/Arduino-Makefile/issues/59))
 - More than one .ino or .pde file is not supported yet ([issue #49](https://github.com/sudar/Arduino-Makefile/issues/49))
 - When you compile for the first time, it builds all libs inside Arduino directory even if it is not needed. But while linking only the relevant files are linked. ([issue #29](https://github.com/sudar/Arduino-Makefile/issues/29)). Even Arduino IDE does the same thing though.
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -261,7 +261,9 @@ else
 endif
 
 # include Common.mk now we know where it is
-include $(ARDMK_DIR)/Common.mk
+ifndef COMMON_INCLUDED
+    include $(ARDMK_DIR)/Common.mk
+endif
 
 # show_config_variable macro is available now. So let's print config details for ARDMK_DIR
 ifndef ARDMK_DIR_MSG
@@ -317,11 +319,7 @@ else
     ifeq ($(ARCHITECTURE),avr)
         ARDUINO_ARCH_FLAG = -DARDUINO_ARCH_AVR
     else
-        ifeq ($(ARCHITECTURE),sam)
-            ARDUINO_ARCH_FLAG = -DARDUINO_ARCH_SAM
-        else
-            ARDUINO_ARCH_FLAG = -DARDUINO_ARCH_$(shell echo $(ARCHITECTURE) | tr '[:lower:]' '[:upper:]')
-        endif
+        ARDUINO_ARCH_FLAG = -DARDUINO_ARCH_$(shell echo $(ARCHITECTURE) | tr '[:lower:]' '[:upper:]')
     endif
 endif
 
@@ -454,7 +452,14 @@ ifndef AVR_TOOLS_DIR
             AVR_TOOLS_DIR = $(SYSTEMPATH_AVR_TOOLS_DIR)
             $(call show_config_variable,AVR_TOOLS_DIR,[AUTODETECTED],(found in $$PATH))
         else
-            echo $(error No AVR tools directory found)
+            # One last attempt using avr-gcc in case using arm
+            SYSTEMPATH_AVR_TOOLS_DIR := $(call dir_if_exists,$(abspath $(dir $(shell which $(avr-gcc)))/..))
+            ifdef SYSTEMPATH_AVR_TOOLS_DIR
+                AVR_TOOLS_DIR = $(SYSTEMPATH_AVR_TOOLS_DIR)
+                $(call show_config_variable,AVR_TOOLS_DIR,[AUTODETECTED],(found in $$PATH))
+            else
+                echo $(error No AVR tools directory found)
+            endif
         endif # SYSTEMPATH_AVR_TOOLS_DIR
 
     endif # BUNDLED_AVR_TOOLS_DIR
@@ -644,6 +649,9 @@ ifeq ($(strip $(NO_CORE)),)
                 USB_PID := $(call PARSE_BOARD,$(BOARD_TAG),menu.(chip|cpu).$(BOARD_SUB).build.pid)
             endif
         endif
+
+        # add caterina flag to ARD_RESET_OPTS
+        ARD_RESET_OPTS += --caterina
     endif
 
     # normal programming info
@@ -745,34 +753,21 @@ endif
 # Reset
 
 ifndef RESET_CMD
-	ARD_RESET_ARDUINO := $(shell which ard-reset-arduino 2> /dev/null)
-	ifndef ARD_RESET_ARDUINO
-		# same level as *.mk in bin directory when checked out from git
-		# or in $PATH when packaged
-		ARD_RESET_ARDUINO = $(ARDMK_DIR)/bin/ard-reset-arduino
-	endif
-    ifneq ($(CATERINA),)
-        ifneq (,$(findstring CYGWIN,$(shell uname -s)))
-          # confirm user is using default cygwin unix Python (which uses ttySx) and not Windows Python (which uses COMx)
-          ifeq ($(shell which python),/usr/bin/python)
-            RESET_CMD = $(ARD_RESET_ARDUINO) --caterina $(ARD_RESET_OPTS) $(DEVICE_PATH)
-          else
-            RESET_CMD = $(ARD_RESET_ARDUINO) --caterina $(ARD_RESET_OPTS) $(call get_monitor_port)
-          endif
-        else
-            RESET_CMD = $(ARD_RESET_ARDUINO) --caterina $(ARD_RESET_OPTS) $(call get_monitor_port)
-        endif
+  ARD_RESET_ARDUINO := $(shell which ard-reset-arduino 2> /dev/null)
+  ifndef ARD_RESET_ARDUINO
+    # same level as *.mk in bin directory when checked out from git
+    # or in $PATH when packaged
+    ARD_RESET_ARDUINO = $(ARDMK_DIR)/bin/ard-reset-arduino
+  endif
+  ifneq (,$(findstring CYGWIN,$(shell uname -s)))
+      # confirm user is using default cygwin unix Python (which uses ttySx) and not Windows Python (which uses COMx)
+      ifeq ($(shell which python),/usr/bin/python)
+        RESET_CMD = $(ARD_RESET_ARDUINO) $(ARD_RESET_OPTS) $(DEVICE_PATH)
+      else
+        RESET_CMD = $(ARD_RESET_ARDUINO) $(ARD_RESET_OPTS) $(call get_monitor_port)
+      endif
     else
-        ifneq (,$(findstring CYGWIN,$(shell uname -s)))
-          # confirm user is using default cygwin unix Python (which uses ttySx) and not Windows Python (which uses COMx)
-          ifeq ($(shell which python),/usr/bin/python)
-            RESET_CMD = $(ARD_RESET_ARDUINO) $(ARD_RESET_OPTS) $(DEVICE_PATH)
-          else
-            RESET_CMD = $(ARD_RESET_ARDUINO) $(ARD_RESET_OPTS) $(call get_monitor_port)
-          endif
-        else
-            RESET_CMD = $(ARD_RESET_ARDUINO) $(ARD_RESET_OPTS) $(call get_monitor_port)
-        endif
+        RESET_CMD = $(ARD_RESET_ARDUINO) $(ARD_RESET_OPTS) $(call get_monitor_port)
     endif
 endif
 
@@ -815,7 +810,7 @@ ifeq ($(strip $(CHK_SOURCES)),)
                 $(call show_config_info,No .pde or .ino files found. If you are compiling .c or .cpp files then you need to explicitly include Arduino header files)
             else
                 #TODO: Support more than one file. https://github.com/sudar/Arduino-Makefile/issues/49
-                $(error Need exactly one .pde or .ino file. This makefile doesn't support multiple .ino/.pde files yet)
+                $(error Need exactly one .pde or .ino file. This makefile doesn\'t support multiple .ino/.pde files yet)
             endif
         endif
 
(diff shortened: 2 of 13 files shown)


sej7278:
egrep instead of grep -E may work, slightly different things

tuna-f1sh:
Yes adding a test would be good, I wasn't sure of the operation so left it out. Do the non-test examples do nothing at the moment. How easy is it to add arm toolchain to travis?

I tried egrep to no avail, same thing. It might even [be a bug](https://unix.stackexchange.com/questions/352977/why-does-this-bsd-grep-result-differ-from-gnu-grep).

tuna-f1sh:
Yes, not sure why I had the word-boundary as works fine without. I'm not that well versed in regex and was toying around with a build tool to create the expression. I looked at using `sed` or `awk` but again, not familiar enough with the tools to use regex with the piped output. Would be good if you can come up with something using them.

tuna-f1sh:
Removing the word-boundary has the same result on both my Arch Linux and macOS High Sierra machine so I've fixed that - thanks.

ShinyMark:
Hi - I'm eager to see this merged in so I can take advantage of this for my new MKR1000. On my Mega I've really been enjoying using Arduino-Makefile to build via the command line in Linux instead of using the Arduino IDE.

I'm an experienced engineer but barely know anything about make as I've been in Windows land for most of my career. I'm happy to try to help test this PR out if it's useful.

sudar:
@ShinyMark 

> I'm happy to try to help test this PR out if it's useful.

That would be very helpful.

I don't have the necessary hardware and so it kind of makes it difficult for me to test it.

tuna-f1sh:
More testing and contributions would be useful. At the moment it's just used by me as far as I know. 

I've tested it on macOS, Linux and Windows and made a few tweaks as the commit log now shows since opening this PR. I think it's now it quite a good state. I have done my work using an Arduino M0 Pro and custom board with M0 Pro bootloader but have not used MK1000.

The thing I have been trying to get working recently is programming using the 'Native' USB. The bootloader emulates an atmega2560 when power cycled with the reset button or opening/closing the serial port at 1200 baud. One can then flash using `avrdude`. The problem I have is that it `avrdude` timeouts during the writing for some reason. It works perfectly in the Arduino IDE but copying even the exact same line into a terminal results in different behaviour! 🤷‍♂️

See video: https://youtu.be/NvkajAe0ikY

Does the same in Arch Linux.

ShinyMark:
Gave this a try today. After some work updating the IDE to the latest (the Ubuntu apt package is ancient) and adding the board support via the IDE I was able do a simple test from the IDE. So running code from the IDE works. Now, on to command line via make.

I tried the `tuna-f1sh:sam` branch. After simple Makefile config for my file system and clarifying my own confusion about the `~/.arduino15` folder I got the MZeroBlink example to compile for the mkr1000 board tag. Unfortunately, `make upload` does not work on my machine. It's not clear to me why.

Here's the error output:

```
/usr/bin/ard-reset-arduino --zero /dev/ttyACM0
usage: ard-reset-arduino [-h] [--caterina] [--verbose] [--period PERIOD] port
ard-reset-arduino: error: unrecognized arguments: --zero
/home/mark/projects/arduino/Arduino-Makefile/Arduino.mk:1630: recipe for target 'reset' failed
```
Is `ard-reset-arduino` the wrong version? If yes, where can I get the right version?

tuna-f1sh:
It's using the installed `ard-reset-arduino` (presumably from a package manager), rather than the branch specific version.

A work around whilst testing it to define `ARD_RESET_ARDUINO = $(ARDMK_DIR)/bin/ard-reset-arduino`

Good to hear it's compiling 👍

ShinyMark:
I put this in an edit to my previous comment but to reiterate for clarity - after uninstalling the `arduino-mk` package I had previously installed via apt everything is working, including upload to my MKR1000. Furthermore, I tested some existing projects of mine on my Mega 2560 from this branch and it also worked.

So it's all looking good to me. Thanks for extending this project for the new boards!

tuna-f1sh:
Good to hear. I've tweaked the reset script slightly so it behaves more like the IDE (polls attached devices until bootloader port attaches or timeout occurs). Seems quite robust now. It's been a trail getting my head around all the SAMD implementations but seems to be getting there now.

tuna-f1sh:
I've done a rebase on this branch as it was getting a bit unwieldy since the original PR. I believe it to be in quite a good state now.

sudar:
@sej7278 What are your thoughts on this PR?

sej7278:
had a quick play and noticed this is hardcoded:

```make
        $(call dir_if_exists, /home/john/.arduino15/packages ) \
```

also it doesn't work with the sparkfun core (the only one i have installed) as it lives inside `~/sketchbook/hardware/` not `~/.arduino15/packages/` in fact `ARDUINO_PACKAGE_DIR` seems quite overused - we've always tended towards supporting the hardware directory not boards manager.

sej7278:
This fixes a few of the issues - Common.mk included earlier, user Makefile should include Sam.mk not Arduino.mk (which then includes Sam.mk!), ARCHITECTURE/ARDMK_VENDOR/VENDOR_NAME substituted.

examples need changing to reflect the above, so does documentation, but then it'll be more like the teensy/chipkit variants.

```diff
diff --git a/Arduino.mk b/Arduino.mk
index 4bd1f7e..842bbfc 100644
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -263,13 +263,6 @@ endif
 # include Common.mk now we know where it is
 include $(ARDMK_DIR)/Common.mk
 
-# include Sam.mk if using SAM or SAMD devices
-ifeq ($(findstring sam, $(strip $(ARDMK_VENDOR))), sam)
-    ifndef SAM_MK_INCLUDED
-        include $(ARDMK_DIR)/Sam.mk
-    endif
-endif
-
 # show_config_variable macro is available now. So let's print config details for ARDMK_DIR
 ifndef ARDMK_DIR_MSG
     $(call show_config_variable,ARDMK_DIR,[COMPUTED],(relative to $(notdir $(lastword $(MAKEFILE_LIST)))))
@@ -834,7 +827,7 @@ ifeq ($(strip $(NO_CORE)),)
         CORE_AS_SRCS    = $(wildcard $(ARDUINO_CORE_PATH)/*.S)
 
 				# USB Core if samd or sam
-				ifeq ($(findstring sam, $(strip $(ARDMK_VENDOR))), sam)
+				ifeq ($(findstring sam, $(strip $(ARCHITECTURE))), sam)
 						CORE_C_SRCS    += $(wildcard $(ARDUINO_CORE_PATH)/USB/*.c)
 						CORE_CPP_SRCS  += $(wildcard $(ARDUINO_CORE_PATH)/USB/*.cpp)
 				endif
@@ -1480,7 +1473,7 @@ AVRDUDE_ISP_OPTS = -c $(ISP_PROG) -b $(AVRDUDE_ISP_BAUDRATE)
 ifndef ISP_PORT
     ifneq ($(strip $(ISP_PROG)),$(filter $(ISP_PROG), atmelice_isp usbasp usbtiny gpio linuxgpio avrispmkii dragon_isp dragon_dw))
         # switch for sam devices as bootloader will be on usb serial if using stk500_v2
-        ifeq ($(findstring sam, $(strip $(ARDMK_VENDOR))), sam)
+        ifeq ($(findstring sam, $(strip $(ARCHITECTURE))), sam)
             AVRDUDE_ISP_OPTS += -P $(call get_monitor_port)
         else
             AVRDUDE_ISP_OPTS += -P $(call get_isp_port)
@@ -1562,7 +1555,7 @@ pre-build:
 # copied from arduino with start-group, end-group
 $(TARGET_ELF): 	$(LOCAL_OBJS) $(CORE_LIB) $(OTHER_OBJS)
 # sam devices need start and end group
-ifeq ($(findstring sam, $(strip $(ARDMK_VENDOR))), sam)
+ifeq ($(findstring sam, $(strip $(ARCHITECTURE))), sam)
 		$(CC) $(LINKER_SCRIPTS) -Wl,-Map=$(OBJDIR)/$(TARGET).map -o $@ $(LOCAL_OBJS) $(OTHER_OBJS) $(OTHER_LIBS) $(LDFLAGS) $(CORE_LIB) -Wl,--end-group
 # otherwise traditional
 else
@@ -1578,7 +1571,7 @@ error_on_caterina:
 # Use submake so we can guarantee the reset happens
 # before the upload, even with make -j
 upload:		$(TARGET_HEX) verify_size
-ifeq ($(findstring sam, $(strip $(ARDMK_VENDOR))), sam)
+ifeq ($(findstring sam, $(strip $(ARCHITECTURE))), sam)
 # do reset toggle at 1200 BAUD to enter bootloader if using avrdude or bossa
 ifeq ($(strip $(UPLOAD_TOOL)), avrdude)
 		$(MAKE) reset
@@ -1592,7 +1585,7 @@ else
 endif
 
 raw_upload:	$(TARGET_HEX) verify_size
-ifeq ($(findstring sam, $(strip $(ARDMK_VENDOR))), sam)
+ifeq ($(findstring sam, $(strip $(ARCHITECTURE))), sam)
 		$(MAKE) do_sam_upload
 else
 		$(MAKE) error_on_caterina
@@ -1648,7 +1641,7 @@ ispload:	$(TARGET_EEP) $(TARGET_HEX) verify_size
 			$(AVRDUDE_ISPLOAD_OPTS)
 
 burn_bootloader:
-ifeq ($(findstring sam, $(strip $(ARDMK_VENDOR))), sam)
+ifeq ($(findstring sam, $(strip $(ARCHITECTURE))), sam)
     ifeq ($(strip $(BOOTLOADER_UPLOAD_TOOL)), openocd)
 				$(OPENOCD) $(OPENOCD_OPTS) -c "telnet_port disabled; init; halt; $(BOOTLOADER_UNPROTECT); program {{$(BOOTLOADER_PARENT)/$(BOOTLOADER_FILE)}} verify reset; shutdown"
     else
diff --git a/Sam.mk b/Sam.mk
index fe5416f..6936986 100644
--- a/Sam.mk
+++ b/Sam.mk
@@ -9,63 +9,57 @@
 #
 ########################################################################
 
-# Flag to say this has been run
-SAM_MK_INCLUDED := 1
-
 ifndef ARDMK_DIR
     ARDMK_DIR := $(realpath $(dir $(realpath $(lastword $(MAKEFILE_LIST)))))
 endif
 
+# include Common.mk now we know where it is
+include $(ARDMK_DIR)/Common.mk
+
 ifndef ARDUINO_PACKAGE_DIR
     # attempt to find based on Linux, macOS and Windows default
     ARDUINO_PACKAGE_DIR := $(firstword \
-        $(call dir_if_exists, /home/john/.arduino15/packages ) \
+        $(call dir_if_exists,$(HOME)/.arduino15/packages) \
         $(call dir_if_exists,$(ARDUINO_DIR)/packages) \
         $(call dir_if_exists,$(HOME)/Library/Arduino15/packages) \
         $(call dir_if_exists,$(USERPROFILE)/AppData/Local/Arduino15/packages) )
 endif
 
 ifndef ARDMK_VENDOR
-    ARDMK_VENDOR := samd
+    ARDMK_VENDOR := arduino
 endif
 
-ifndef VENDOR_NAME
-    VENDOR_NAME := arduino
+ifndef ARCHITECTURE
+    ARCHITECTURE := samd
 endif
 
 ifndef CORE_VER
-    CORE_VER := $(shell basename $(wildcard $(ARDUINO_PACKAGE_DIR)/$(VENDOR_NAME)/hardware/$(ARDMK_VENDOR)/1.*))
+    CORE_VER := $(shell basename $(wildcard $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/hardware/$(ARCHITECTURE)/1.*))
 endif
 
 ifndef CMSIS_VER
-    CMSIS_VER := $(shell basename $(wildcard $(ARDUINO_PACKAGE_DIR)/$(VENDOR_NAME)/tools/CMSIS/4.*))
+    CMSIS_VER := $(shell basename $(wildcard $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/tools/CMSIS/4.*))
 endif
 
 ifndef CMSIS_ATMEL_VER
-    CMSIS_ATMEL_VER := $(shell basename $(wildcard $(ARDUINO_PACKAGE_DIR)/$(VENDOR_NAME)/tools/CMSIS-Atmel/1.*))
+    CMSIS_ATMEL_VER := $(shell basename $(wildcard $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/tools/CMSIS-Atmel/1.*))
 endif
 
 ifndef CMSIS_DIR
-    CMSIS_DIR := $(ARDUINO_PACKAGE_DIR)/$(VENDOR_NAME)/tools/CMSIS/$(CMSIS_VER)/CMSIS
+    CMSIS_DIR := $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/tools/CMSIS/$(CMSIS_VER)/CMSIS
 endif
 
 ifndef CMSIS_ATMEL_DIR
-    CMSIS_ATMEL_DIR := $(ARDUINO_PACKAGE_DIR)/$(VENDOR_NAME)/tools/CMSIS-Atmel/$(CMSIS_ATMEL_VER)/CMSIS
-endif
-
-# include Common.mk now we know where it is and hasn't been called already
-ifndef show_separator
-    include $(ARDMK_DIR)/Common.mk
+    CMSIS_ATMEL_DIR := $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/tools/CMSIS-Atmel/$(CMSIS_ATMEL_VER)/CMSIS
 endif
 
 # Arduino Settings
-ARCHITECTURE        = arm
-ARDUINO_CORE_PATH   = $(ARDUINO_PACKAGE_DIR)/$(VENDOR_NAME)/hardware/$(ARDMK_VENDOR)/$(CORE_VER)/cores/arduino
-ALTERNATE_CORE_PATH = $(ARDUINO_PACKAGE_DIR)/$(VENDOR_NAME)/hardware/$(ARDMK_VENDOR)/$(CORE_VER)
-BOARDS_TXT          = $(ARDUINO_PACKAGE_DIR)/$(VENDOR_NAME)/hardware/$(ARDMK_VENDOR)/$(CORE_VER)/boards.txt
+ARDUINO_CORE_PATH   = $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/hardware/$(ARCHITECTURE)/$(CORE_VER)/cores/arduino
+ALTERNATE_CORE_PATH = $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/hardware/$(ARCHITECTURE)/$(CORE_VER)
+BOARDS_TXT          = $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/hardware/$(ARCHITECTURE)/$(CORE_VER)/boards.txt
 
 # Check boards file exists before continuing as parsing non-existant file can create problems
-ifneq ($(findstring boards.txt, $(wildcard $(ARDUINO_PACKAGE_DIR)/$(VENDOR_NAME)/hardware/$(ARDMK_VENDOR)/$(CORE_VER)/*.txt)), boards.txt)
+ifneq ($(findstring boards.txt, $(wildcard $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/hardware/$(ARCHITECTURE)/$(CORE_VER)/*.txt)), boards.txt)
     echo $(error $(CORE_VER) Cannot find boards file $(BOARDS_TXT). Check ARDUINO_PACKAGE_DIR path: $(ARDUINO_PACKAGE_DIR) and board support installed)
 endif
 
@@ -95,12 +89,12 @@ ALT_CORE_S_SRCS := $(wildcard $(ALTERNATE_CORE_PATH)/variants/$(VARIANT)/*.S)
 
 # Use arm-toolchain from Arduino install if exists and user has not defined global version
 ifndef ARM_TOOLS_DIR
-    ARM_TOOLS_DIR = $(call dir_if_exists,$(wildcard $(ARDUINO_PACKAGE_DIR)/$(VENDOR_NAME)/tools/arm-none-eabi-gcc/*))
+    ARM_TOOLS_DIR = $(call dir_if_exists,$(wildcard $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/tools/arm-none-eabi-gcc/*))
 endif
 
 # define plaform lib dir ignoring teensy's oversight on putting it all in avr
 ifndef ARDUINO_PLATFORM_LIB_PATH
-    ARDUINO_PLATFORM_LIB_PATH := $(ARDUINO_PACKAGE_DIR)/$(VENDOR_NAME)/hardware/$(ARDMK_VENDOR)/$(CORE_VER)/libraries
+    ARDUINO_PLATFORM_LIB_PATH := $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/hardware/$(ARCHITECTURE)/$(CORE_VER)/libraries
     $(call show_config_variable,ARDUINO_PLATFORM_LIB_PATH,[COMPUTED],(from ARDUINO_PACKAGE_DIR))
 endif
 
```

tuna-f1sh:
Thank you for taking the time to look through this and create the patch. Changes make sense to me, I'll test them throughout the week.

I think my thinking with using ARDMK_VENDOR rather than ARCHITECTURE was due to the ARCH define and becoming confused with Arduino originally using AVR correctly as an architecture but now using a chip series instead! I agree that it is best not to deviate from this legacy though.

The folder structure of Arduino seems less than ideal but I guess we have to work with it. I found it confusing how the auxiliary package hardware path is 'hardware/arduino/avr...' but packages are 'arduino/hardware/sam...'. I assume if one sets ARDUINO_PACKAGE_DIR as 'sketchbook/hardware/' this would still resolve correctly? The CORE_VER should work if there is not core version sub-directory as the wildcard CORE_VER will not get defined (I tested this using the Arch linux manual install of the samd support package). 

Those are my first thoughts but I'll work in the changes and feedback here. I would recommend getting a dev board with both the on-board CMSIS_DAP programmer port and native usb port, so that you can test different bootloaders - unless you have an external CMSIS debugger already. They are quite a bit more expensive however, as I don't think anyone has [reverse engineered Atmels EDBG firmware](https://www.eevblog.com/forum/microcontrollers/anyone-poked-around-the-atmel-edbg/) yet. I think you could use a ST-LINK jtag clone with OpenOCD however. That zero clone you linked is probably a good starting point though.

sej7278:
ah i never thought of setting the package dir to hardware, i tried just manually setting alternate_core_path or somesuch. i did have some problems with the core_ver checks then (as there is no version in sparkfun's "hardware dir" core).

thanks for the advice re devboard. i've got an st-link from my stm32/stm8 days. i wonder if anyone has done a proper arduino core for the spark core/photon, rather than particle's own wiring port....

tuna-f1sh:
I'm not sure. If you're willing to spend a bit more, this is the board I use. It is essentially an Arduino M0 Pro but in a different - but potentially more useful - form factor and cheaper: https://uk.rs-online.com/web/p/processor-microcontroller-development-kits/1306174/

sej7278:
i like the new changes - one typo is you're using the string "CORE_VER" instead of the variable $CORE_VER

```diff
diff --git a/Sam.mk b/Sam.mk
index 1655e60..23442e6 100644
--- a/Sam.mk
+++ b/Sam.mk
@@ -51,7 +51,7 @@ endif
 ifndef CORE_VER
     CORE_VER := $(wildcard $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/hardware/$(ARCHITECTURE)/1.*)
     ifdef CORE_VER
-        CORE_VER := $(shell basename CORE_VER)
+        CORE_VER := $(shell basename $(CORE_VER))
          $(call show_config_variable,CORE_VER,[DETECTED],(from ARDUINO_PACKAGE_DIR))
     endif
 else
```

also seem to be seeing some duplication of the show_config_variable output

tuna-f1sh:
Thanks. It's still a bit work in progress implementing those changes, just on two machines and wanted to keep progress - probably should have put on another branch. Good spot on the CORE_VER variable.

The `show_config_variable` was something I was going to ask your opinion on. I remembered it was one of the reasons I went down the road of defining ARDMK_VENDOR as samd then putting up Sam.mk via Arduino.mk rather than the user including Sam.mk directly in their Makefile - so that the config variables printed with the output. The header guard sounds like a good solution.

I'm going to move the bossac stuff to above the `show_separator` - it fitted nicely with the avrdude opt stuff but I don't want to move `show_separator` further down the file. 

I'll make the relevant changes to the supporting docs too considering these changes from my original implementation.

ShinyMark:
After pulling latest my local MKR1000 test no longer builds due to what appears to be an inability to find the samd boards.txt. I'm sure it's a configuration problem on my end. An updated MZeroBlink Makefile example would help me understand what needs to be updated given the latest changes.

sej7278:
@ShinyMark there's a typo in Sam.mk, the fix is to replace `CORE_VER` with `$(CORE_VER)` at line 54:

-        CORE_VER := $(shell basename CORE_VER)
+        CORE_VER := $(shell basename $(CORE_VER))

tuna-f1sh:
I've corrected the typo and other points so it should work for you again now @ShinyMark . You will have to change your project Makefile like the examples Makefile changes (include Sam.mk rather than Arduino.mk and don't define ARDMK_VENDOR).

sej7278:
can we wrap the openocd/bossac output in a sam arch check, as when using Arduino.mk for an uno for instance, its outputting the openocd/bossac paths?

```diff
diff --git a/Arduino.mk b/Arduino.mk
index 48bc5af..c601995 100644
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1196,6 +1196,8 @@ else
     $(call show_config_variable,BOOTLOADER_PARENT,[USER])
 endif
 
+ifeq ($(findstring sam, $(strip $(ARCHITECTURE))), sam)
+
 ########################################################################
 # OpenOCD for SAM devices
 
@@ -1252,6 +1254,8 @@ else
     BOSSA_OPTS += --port=$(ISP_PORT)
 endif
 
+endif
+
 ########################################################################
 # Tools version info
 ARDMK_VERSION = 1.6
```

tuna-f1sh:
I moved the defines to Sam.mk since it works there and makes sense since the ARM defines are there too. Not sure what you mean about not being able to test upload works? 

I also changed the header output to change the filename depending on the calling file and added a separator at the end of Sam.mk. That way it is easy to differentiate between defines in each file

sej7278:
haha, i had thought of splitting the header output to show each *.mk file too! nice.

i meant i can't test of bossac/openocd can actually upload as i don't have a board.

seems pretty much good to go to me :+1:

README.md
```diff
@@ -313,6 +315,59 @@ See examples/BlinkOpenCM for example usage.
...
+Native USB programing using Bossa (Zero, MKR1000, Feather style bootloaders)
+and avrdude (M0 bootloaders) is supported. The bootloaders on these devices
+requires a double press of the reset button or open/closing the serial port at
+1200 BAUD. The automatic entry of the bootloader is attempted using
+`ard-reset-arduino` when using the general `make upload` target by polling
+attached devices until the bootloader port re-attaches (same method as the
+IDE). On Windows, the USB enumerates as differnt COM ports for the CDC Serial
+and bootloader and these must be defined. On encountering problems, one can
+manually enter the bootloader then upload using the `make raw_upload` target.
+Note youthat the `make reset` target will enter the bootloader on these devices;
```
sej7278: youthat typo

tuna-f1sh:
Thanks. I tried this with the sparkfun core too and noticed the PLATFORM_LIB_PATH should use the ALTERNATIVE_CORE_PATH. I've added some notes in the read me section too.

sej7278:
yes one mega patch would be better than 8 commits please.

i think we're good-to-go if you're happy?

@sudar any issues (once its rebased)?

ShinyMark:
I grabbed latest again and am hitting an error on the MKR1000. Compilation works but I am unable to upload to the device. I confirmed that uploading via the IDE works. Here's the last bit of the make log:

```
mkdir -p build-mkr1000
make reset
make[1]: Entering directory '/home/mark/projects/arduino/samtest'
~/projects/arduino/Arduino-Makefile/bin/ard-reset-arduino --zero /dev/ttyACM0
/dev/ttyACM0
make[1]: Leaving directory '/home/mark/projects/arduino/samtest'
make do_sam_upload
make[1]: Entering directory '/home/mark/projects/arduino/samtest'
mkdir -p build-mkr1000
/usr/bin/bossac -d --info --erase --write --verify --reset --port=/dev/ttyACM0 build-mkr1000/samtest.bin
No device found on /dev/ttyACM0
```

The device is definitely on /dev/ttyACM0. So I'm not sure what's wrong here.

tuna-f1sh:
Hmm, that is odd that it can't find the device. Could it be a permissions thing on the port? It looks like it isn't entering the bootloader (unless it is already in the bootloader) as the port isn't changing after the reset. Couple of things to try:

* What happens if you manually enter the bootloader by pressing the reset button twice then using the 'make raw_upload' target?
* Can you capture the output with `ARD_RESET_OPTS = --verbose`. You'll have to pull the code as I noticed ARD_RESET_OPTS was set rather than appending.

sudar:
@sej7278 

Let's merge it once it's been rebased into a single commit.

@tuna-f1sh Thanks for your contributions and patience.

tuna-f1sh:
@sudar  Thanks. I'll rebase to a single commit once we've got to the bottom of @ShinyMark's problem.

sej7278:
sounds good, we could do with merging #546 and #533 after, they might need rebasing though

tuna-f1sh:
I think I worked out your problem @ShinyMark. I think it is the change to use upload tool on path over the Arduino supplied tool if it is available. Your terminal output is calling `/usr/bin/bossac` so I assume you have installed it via your package manager. 

You distribution must have an older version of bossa (or you installation is out of date) as the original Arduino support was in another branch. V 1.8 is what I've tested with and it works (I believe 1.7 also incorporates the changes). You can either upgrade your bossa or define `BOSSA = $(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/tools/bossac/1.7.0/bossac`

ShinyMark:
Confirmed, that was the problem and your suggestion fixed it. The `bossac` in my path is version 1.3a. I'm not sure how or when that was installed. If that can be detected and have an error printed that might be helpful for others.

Thanks for the help!

sej7278:
the path to bossac is already printed. seems debian is 1.3a, it doesn't seem to have a simple -V option or whatever to get the version.

tuna-f1sh:
I've changed the detection so that it is similar to the avr tools method: check support package path for the tool first, then fall back to user path. Since the Arduino supplied should always be the right version this seems sensible. Arch has V 1.8 since it likes to be bleeding edge but I have a feeling most other distros have the older version that doesn't have the SAMD21 support.

I'll add a note somewhere in the README about this too.

tuna-f1sh:
I've rebased this into one commit now so unless I've made any mistakes in the last fix, it should be good to go 👍

PR #546, opened 2018-01-21
Allow compatibility with old pgmspace API to be disabled by the user

kierdavis:
#169 added `-D__PROG_TYPES_COMPAT__` to the compiler flags, which enables compatibility with the old PROGMEM API (in which `PROGMEM` attributes were attached to types rather than variables). Usage of the old API raises compilation warnings about its deprecation, which is good.

Due to the fact that some symbols (`PGM_P` and `PGM_VOID_P` in particular) are common to both the old and new versions of the API but with differing implementations, code that uses only the new API can still raise bogus deprecation warnings if `__PROG_TYPES_COMPAT__` happens to be enabled.

Moreover, it would be better to encourage users to upgrade outdated code to the new interface than to maintain eternal compatibility with a deprecated API - the user should have to explicitly declare that they want to enable compatibility mode.

Therefore I propose that the default behaviour should be changed to leave `__PROG_TYPES_COMPAT__` unset. The user has the ability to re-enable it if they like, by adding

```makefile
CPPFLAGS += -D__PROG_TYPES_COMPAT__
```

to their Makefile.

Relatedly, is there a good place we could document this? I feel a FAQ-type section featuring the error message(s) that may require enabling `__PROG_TYPES_COMPAT__` would be a good fit for this, but I couldn't see a could place to put it.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1032,10 +1032,16 @@ else
 endif
 
 # Using += instead of =, so that CPPFLAGS can be set per sketch level
-CPPFLAGS      += -$(MCU_FLAG_NAME)=$(MCU) -DF_CPU=$(F_CPU) -DARDUINO=$(ARDUINO_VERSION) $(ARDUINO_ARCH_FLAG) -D__PROG_TYPES_COMPAT__ \
+CPPFLAGS      += -$(MCU_FLAG_NAME)=$(MCU) -DF_CPU=$(F_CPU) -DARDUINO=$(ARDUINO_VERSION) $(ARDUINO_ARCH_FLAG) \
         -I$(ARDUINO_CORE_PATH) -I$(ARDUINO_VAR_PATH)/$(VARIANT) \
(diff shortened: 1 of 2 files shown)


sej7278:
if you removed the flag from Arduino.mk it could be re-enabled by putting the following in the sketch and not changing the local Makefile settings:

```cpp
#ifndef __PROG_TYPES_COMPAT__
#define __PROG_TYPES_COMPAT__
#endif
```

but the 1.8.5 ide still calls `-D__PROG_TYPES_COMPAT__` and we tend to stick with that so as not to cause issues where the ide would build something but the makefile won't.

kierdavis:
Regarding staying compatibly with the IDE, that's totally reasonable.

Therefore, I've changed the direction of the PR somewhat to leave `__PROG_TYPES_COMPAT__` enabled by default, but to change the position in which it's placed in the `CPPFLAGS` variable to allow it to be overriden by a user if they wish.

I feel this is a good compromise. What are your thoughts?

kierdavis:
Thanks :+1:

sej7278:
thanks for the PR

PR #556, opened 2018-03-18
Added the DIAGNOSTICS_COLOR_WHEN var for controlling diagnostics-color.

wingunder:
Some IDEs do not work nicely with the recently added diagnostics-color
compiler switch. This patch adds the DIAGNOSTICS_COLOR_WHEN variable,
for controlling diagnostics-color switch. The default was chosen to be
'always', which makes the Arduino.mk file backwards compatible with
everything up to commit fa82c3a9db (Sat Sep 2 15:32:52 2017 +0100).

The supported values for DIAGNOSTICS_COLOR_WHEN are: `always`, `never`
and `auto`, and as with all Arduino.mk variables, it can easily be
overridden. For more details on the diagnostics-color compiler switch,
see: https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/Language-Independent-Options.html#Language-Independent-Options

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1108,10 +1108,11 @@ endif
 CFLAGS        += $(CFLAGS_STD)
 CXXFLAGS      += -fpermissive -fno-exceptions $(CXXFLAGS_STD)
 ASFLAGS       += -x assembler-with-cpp
+DIAGNOSTICS_COLOR_WHEN ?= always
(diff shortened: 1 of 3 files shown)


sudar:
@wingunder 

wingunder:
Thanks :+1:

PR #564, opened 2018-04-29
Change := to += to allow projects to add their own objs to OTHER_OBJS

myelin:
OTHER_OBJS is an undocumented, but useful, way to add extra sources to
a project.  Setting it with := here breaks this functionality, so I've
changed it to += to not trample on the user-set value.


sudar:
@myelin 

Thanks for the PR.

Unfortunately I didn't get to it on time and it has resulted in some merge conflicts.

Can you please update your branch and fix the merge conflict?

myelin:
It looks like this has been obsoleted by #567, which removed this code and stopped overwriting OTHER_OBJS.  Closing as unnecessary :)

PR #573, opened 2018-07-25
Arduino.mk: Fix clean target declaration

rzr:
Write clean: instead of clean::

This is needed on some systems

Change-Id: Ie9186b7c5b6920327846e42483f0299d42ce693b
Signed-off-by: Philippe Coval <p.coval@samsung.com>

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1663,7 +1663,7 @@ ifneq ($(strip $(AVRDUDE_ISP_FUSES_POST)),)
 		$(AVRDUDE) $(AVRDUDE_COM_OPTS) $(AVRDUDE_ISP_OPTS) $(AVRDUDE_ISP_FUSES_POST)
 endif
 
-clean::
+clean:
 		$(REMOVE) $(OBJDIR)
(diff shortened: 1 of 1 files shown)


sudar:
@rzr 

Thanks for your Pull request.

@sej7278 

Do you remember why we have an extra `:` for `clean`? Was it a typo?

sej7278:
seems its pretty normal for make clean to have double colons, don't quite understand why but i'd rather not remove it without more justification than "This is needed on some systems" ;-)

https://www.gnu.org/software/make/manual/html_node/Double_002dColon.html

sudar:
@sej7278 

Thanks.

@rzr 

Can you please give us more information about the exact issue that you have been facing because of the double colon?

rzr:
I know that I had to change it , but I don't remember why...

sudar:
@wingunder 

Thanks for the explanation. Totally makes sense.

@rzr 

I am closing this without merging for the above reasons. Do re-open if you have more information.

PR #578, opened 2018-08-09
Fix typo of AVRDUDE_ISP_OPT

TE-HiroakiYamazoe:
Currently `make ispload` on Windows does not work because of typo as below.

- AVRDUDE_ISP_OPT should be AVRDUDE_ISP_OPTS
- ISP_PORT should be $(ISP_PORT)

This pull request fixes the typo.


sej7278:
thanks for looking at this

PR #580, opened 2018-09-14
Moved the PARSE_BOARD macro and added the TOOL_PREFIX variable for setting up tool chains.

wingunder:
This pull request consists of 2 commits, as the second commit really required the first commit:

1. Moved the PARSE_BOARD macro from Arduino.mk to Common.mk.                                                                               
                                                                                                                                        
   There seems to be 3 different macros to parse the boards.txt file.                                                                      
   This patch moves the PARSE_BOARD macro from Arduino.mk to Common.mk.                                                                    
   The PARSE_OPENCM and PARSE_TEENSY macros in Teensy.mk and OpenCM.mk                                                                     
were removed and the common PARSE_BOARD is now being called from                                                                        
everywhere.                                                                                                                             
                                                                                                                                        
   Advantages of this fix are:                                                                                                             
      * Less code, i.e. no redundant parse macros.                                                                                         
      * A single standardized algorithm to parse the boards.txt file.                                                                      

2. Added the TOOL_PREFIX variable for setting up the executable tools.                                                                     
                                                                                                                                        
   Currently three different tool chains seem to be used:                                                                                  
      * avr-*                                                                                                                              
      * pic32-*                                                                                                                            
      * arm-none-eabi-*                                                                                                                    
                                                                                                                                        
   These all get set up independently. This patch centralizes the                                                                          
definitions of the executable tools and does it generically, by means                                                                   
of the newly introduced TOOL_PREFIX variable. Setting up a                                                                              
tool chain is now simply a matter of defining the TOOL_PREFIX                                                                           
variable. For the currently supported tool chains it gets set to avr,                                                                   
pic32 or arm-none-eabi. Arbitrary tool chains can now easily be set up,                                                                 
by the TOOL_PREFIX variable.                                                                                                            
                                                                                                                                        
   Although the use of the OVERRIDE_EXECUTABLES variable is now almost                                                                     
not justifiable, it was left as-is, in order to assure backwards                                                                        
compatibility.


sej7278:
there are 2 different parse_board routines, opencm/teensy are the same; and sam/avr are the same, with slightly different regexes for handling whitespace, and levels of cutting on colons/equals, are these merged safely into one?

```make
PARSE_OPENCM = $(shell grep -v "^\#" "$(BOARDS_TXT)" | grep $(1).$(2) | cut -d = -f 2- )
PARSE_TEENSY = $(shell grep -v "^\#" "$(BOARDS_TXT)" | grep $(1).$(2) | cut -d = -f 2- )

PARSE_BOARD = $(shell grep -Ev '^\#' $(BOARDS_TXT) | \
     grep -E '^[ \t]*$(1).$(2)=' | cut -d = -f 2 | cut -d : -f 2)
```

this looks like excellent work though and may even pave the way for adding more toolchains e.g. xtensa

wingunder:
Yes, I was fully aware of the the differences in the different parse algorithms. I'll go through them one by one (mostly for all readers to understand it):

`grep -Ev '^\#'` vs `grep -v '^\#'`: Basically `-E` is just for 'extended', which won't play any role for the regexp `'^\#'`, as it is a basic regexp, filtering the empty lines.

`grep -E '^[ \t]*$(1).$(2)='` vs `grep $(1).$(2)`: The first one _needs_ the -E, as it's a more complicated (extended) regexp. Adding a -E to the second grep statement will have not have any affect. The `^[ \t]*` in the first regexp filters the start of line, _optionally_ followed by tabs. It then filters the term `$(1).$(2)`, like the second regexp, but expects this to be followed by an equals sign. This is forwards compatible with the second regexp, as we'll just throw away any leading tabs of a line in the boards.txt file. The equals sign is in any case mandatory and part of the definition of the boards.txt files. The equal sign filtering has its pros and cons, but for now, lets stick to the basics and keep it as it is.

`cut -d = -f 2` vs `cut -d = -f 2-`: The first one will return the second term, whereas the second one will return all terms from the 2nd term onward. This will only return different results for an attribute that has an equal sign in it, eg:
```
crazyboard.name=duffboard = duff-board
$ echo "crazyboard.name=duffboard = duff-board" |grep -E '^[ \t]*[a-zA-Z\.]+=' |cut -d = -f 2 
duffboard 
$ echo "crazyboard.name=duffboard = duff-board" |grep -E '^[ \t]*[a-zA-Z\.]+=' |cut -d = -f 2-
duffboard = duff-board
```
I have grepped through all the boards.txt files that I could find, and could not find an example like this. 
So for all intents and purposes `cut -d = -f 2` will be ok.

`cut -d = -f 2 | cut -d : -f 2` vs `cut -d = -f 2-`: The first command basically allows a board definition with either `=` or `:` as the key, value separator. This however has no chance of working, as the grep, preceding the first command _requires_ an equals sign. So even this will result in the same, assuming that `cut -d = -f 2` and `cut -d = -f 2-` will return the same.

So, to wrap it up, I tried to keep as much as possible, the same as before. The aim was to make it easy to add new tool chains easily, without fixing all the potential parsing problems. My idea is to make the PARSE_BOARD macro much more robust, but this is material for another patch.

I hope that I could convince you.

wingunder:
Here's some more information on the parsing of values containing colons, using `cut -d : -f 2`:

The only documented keys with values containing colons where the following:

From: https://code.google.com/archive/p/arduino/wikis/Platforms.wiki

> BOARD.build.core: the location of the core to link sketches against when compiling for the board. This may be the name of a folder in the core sub-folder of the platform (e.g. "arduino") or a value of the format platform:core, where platform is the name of the folder of another platform (e.g. the built-in "arduino") and core is a folder in that platform's core folder.

> BOARD.bootloader.path: the location of the folder containing the bootloader .hex file. This may be the name of a sub-folder of the bootloader sub-folder of the board's platform, or a value of the form platform:path, where platform is the name of the folder of another platform, and path is a sub-folder of its bootloaders sub-folder. Note: the bootloader file and path are optional; if omitted, the "burn bootloader" command will simply set the fuses.

Both echo "build.core" and "bootloader.path" ended up with the same parsing rules after this patch, which means that we are parsing the values containing colons correctly, as before. 

In commit e639f81b, the parsing of build.core was fixed. If I take the exact example from this commit, I get the right answer:
```
$ echo "build.core=arduino:arduino" | grep -Ev '^\#'  | grep -E "^[ \t]*build.core=" | cut -d = -f 2 | cut -d : -f 2
arduino
```

It will be much appreciated if anyone can point me to some funky bords.txt files or some examples.
BTW, I know about these ones:
   * /usr/share/arduino/hardware/arduino/boards.txt at https://packages.debian.org/sid/all/arduino-core/filelist
   * https://raw.githubusercontent.com/arduino/arduino-flash-tools/master/hardware/avr/boards.txt
   * https://raw.githubusercontent.com/arduino/ArduinoCore-avr/master/boards.txt
   * https://raw.githubusercontent.com/adafruit/Adafruit_Arduino_Boards/master/boards.txt

Interestingly enough, it seems as if there is a lot of uses of values containing colons: 
```
$ wget --quiet -O - https://raw.githubusercontent.com/adafruit/Adafruit_Arduino_Boards/master/boards.txt |grep -ve ^# -ve ^$ |grep -e : | awk -F'=' '{print $1}' |cut -d . -f2- |awk '{print "board." $0}'|sort |uniq
```
returns:
```
board.bootloader.tool
board.build.core
board.build.variant
board.upload.tool
```
I am however sure that PARSE_BOARD will parse things correctly, as shown in the following example, first with and then without a colon:
```  
$ echo protrinket5ftdi.bootloader.tool=arudino:avrdude |cut -d = -f2 | cut -d : -f2 
avrdude
$ echo protrinket5ftdi.bootloader.tool=avrdude |cut -d = -f2 | cut -d : -f2 
avrdude
```

sej7278:
thanks for the thorough explanation, saves me the time of going through it all.

some awkward boards.txt for you to try, some of these need a BOARD_SUB and don't parse properly now anyway and you have to manually define e.g. F_CPU

https://packages.debian.org/sid/all/arduino-mighty-1284p/filelist
https://github.com/Lauszus/Sanguino/blob/master/boards.txt
https://github.com/SpenceKonde/ATTinyCore/blob/master/avr/boards.txt
https://github.com/MCUdude/MegaCore/blob/master/avr/boards.txt

sej7278:
i can't get your fork to work with any core, it seems to get stuck at:

```
- [AUTODETECTED]       ARDUINO_SKETCHBOOK = /home/simon/arduino16 (from arduino preferences file)
```

and never gets to the next step:

```
- [BUNDLED]            AVR_TOOLS_DIR = /home/simon/arduino-1.8.6/hardware/tools/avr (in Arduino distribution)
```

My 1.8.6 Makefile:
```make
ARDUINO_DIR = /home/simon/arduino-1.8.6
BOARD_TAG = mega
BOARD_SUB = atmega2560
include /home/simon/wingunder/Arduino-Makefile/Arduino.mk
```

or 1.0.5:

```make
BOARD_TAG = mega2560
include /home/simon/wingunder/Arduino-Makefile/Arduino.mk
```

wingunder:
> i can't get your fork to work with any core on debian sid, it seems to get stuck at:

Right, sorry about that, but I saw that as well. The reason is that the auto detection of BOARDS_TXT is not working for Debian, as the file's path is wrong. Could you please echo BOARDS_TXT right after setting ARDUINO_SKETCHBOOK and let me know what you get. 

I have a patch for this, but I'm half way down your boards.txt list and I wanted to finish that, before jumping onto this.

wingunder:
The 62d23d6 patch above won't fix the problems on Debian, but it will at least not 'hang'. If the BOARDS_TXT variable is not pointing to a file, it will stop with an error, stating just that.

wingunder:
> The 62d23d6 patch above won't fix the problems on Debian

The 62d23d6 patch above _might_ fix your problems on Debian.

wingunder:
> The 62d23d6 patch above _might_ fix your problems on Debian.

Your makefile works for me:
``` 
$ pwd
......./Arduino-Makefile/examples/Blink
$ git log |head -1
commit 62d23d66a2f9326235022bfd8711c62e015c0bc8
$ echo "BOARD_TAG = mega
BOARD_SUB = atmega2560
include ../../Arduino.mk
" >Makefile.atmega2560
$ make -f Makefile.atmega2560 >/dev/null 2>&1 ; echo $?
0
$ cat /etc/debian_version
buster/sid
$ uname -a
Linux xxxxxx 4.17.0-3-amd64 #1 SMP Debian 4.17.17-1 (2018-08-18) x86_64 GNU/Linux
```

sej7278:
seems to be working now with debian's 1.0.5 as well as upstream 1.8.6 and various cores

sej7278:
i think its great, @sudar perhaps you could test on macos?

sej7278:
just tested with 1.8.7 and Sam.mk, works fine.

@sudar what do you think about merging this?

sudar:
@sej7278 

Thanks for testing it out. Let's merge it.

PR #583, opened 2018-09-29
Moved CORE_LIB to the last position of the defined linked objects.

wingunder:
The linking order was changed from:
   $(LOCAL_OBJS) $(CORE_LIB) $(OTHER_OBJS) $(OTHER_LIBS)
to:
   $(LOCAL_OBJS) $(OTHER_OBJS) $(OTHER_LIBS) $(CORE_LIB)

This makes more sense, as OTHER_OBJS would rather depend on CORE_LIB,
than the other way around.

Apart from libc and libm, CORE_LIB should conceptually _always_ be
the last lib in the link list, as it is the core (base of everything).
BTW, this was implemented correctly for the 'sam' architecture.


sudar:
@sej7278 

I am interested to know your thoughts on this.

sej7278:
i actually didn't think the order mattered, but it seems ok

wingunder:
> i actually didn't think the order mattered

I am sure you are already convinced, but just so that all understand what's happening here, I'll explain it by example:

If you run make in examples/Blink, the following line will link the code:
`$ avr-gcc -mmcu=atmega328p -Wl,--gc-sections -Os -flto -fuse-linker-plugin -o build-uno/Blink.elf build-uno/Blink.ino.o build-uno/libcore.a   -lc -lm
`

Let's now move libcore.a to just before Blink.ino.o. The Blink.ino.o object _needs_ pinMode and digitalWrite, that resides in libcore.a. Our linker, that links from left to right, now claims that it can't find the definitions of pinMode and digitalWrite:
`
$ avr-gcc -mmcu=atmega328p -Wl,--gc-sections -Os -flto -fuse-linker-plugin -o build-uno/Blink.elf build-uno/libcore.a build-uno/Blink.ino.o  -lc -lm`

```
/usr/lib/gcc/avr/5.4.0/../../../avr/bin/ld: wiring_digital.c.o: plugin needed to handle lto object
/tmp/ccWJhLWH.ltrans0.ltrans.o: In function `main':
<artificial>:(.text.startup+0x8): undefined reference to `pinMode'
<artificial>:(.text.startup+0x14): undefined reference to `digitalWrite'
<artificial>:(.text.startup+0x20): undefined reference to `digitalWrite'
collect2: error: ld returned 1 exit status
```
In we build examples/Blink, things work, as Blink.ino.o is part of LOCAL_OBJS (which appears _before_ CORE_LIB in the link command). Lets assume we link another lib to examples/Blink, as part of OTHER_LIB (which appears _after_ CORE_LIB in the link command). If this 'other' lib now needs _anything_ from libcore.a, the outcome will be similar to the example above and the linking exercise will fail.

wingunder:
It seems as if something is not so healthy with the Travis testing:
```
/usr/bin/sudo: illegal option -- E
This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid, and setgid executables.
If you require sudo, add 'sudo: required' to your .travis.yml

```
 I'm pretty sure it's not the new directory, examples/Blink3rdPartyLib, that I added. I tested everything with `./tests/script/runtests.sh`, before committing and all tests passed, including the compilation in the newly added directory.

Could you please have a look?
Thanks

wingunder:
> It seems as if something is not so healthy with the Travis testing:
> Could you please have a look?

Don't bother, it seems to now be fixed in PR #584.

@sudar: Once PR #584 is merged/fixed, could you please manually kick off a the check for this PR (#583)?
Thanks in advance.

sudar:
@wingunder 

wingunder:
> Thanks for the PR. Just merged it.

@sudar No problem. Thanks for merging it.
@sej7278 Thanks for you reviews.

PR #586, opened 2018-10-13
allow to enable AVRDUDE_AUTOERASE_FLASH

devel0:
I using usbasp to program atmega328p barebone so I disabled ard-reset-arduino by declaring `RESET_CMD = echo` ; btw I got problem flashing image within -D option of avrdude enabled, so I introduced this option AVRDUDE_AUTOERASE_FLASH=yes to allow inhibit (default) deactivation of autoerase flash.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1490,7 +1490,11 @@ endif
 # -D - Disable auto erase for flash memory
 # Note: -D is needed for Mega boards.
 #       (See https://github.com/sudar/Arduino-Makefile/issues/114#issuecomment-25011005)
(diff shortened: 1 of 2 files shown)


sej7278:
you shouldn't need to do any of that, just set ```AVRDUDE_OPTS = -v``` or somesuch to remove the default -D

devel0:
I didn't think so cause -D hardcoded ( see [this](https://github.com/sudar/Arduino-Makefile/blob/d4d3d3d6c354db1e2989d8dd671f8af96cc036d8/Arduino.mk#L1493) )

sej7278:
@sudar i don't think this will hurt being merged, it doesn't change the default behaviour :+1:

sudar:
Thanks @sej7278 just merged it.

Thanks @devel0 for your contribution.

Issue #576, opened 2018-07-28
/usr/etc/avrdude.conf is not the standard path on Arch Linux


alxpettit:
Makefile only looks under `/usr/etc/avrdude.conf`. On Arch Linux, it installs under `/etc/avrdude.conf` (which to me seems more in keeping with standard directory paths on modern Linux systems).

This does not affect functionality in the Arduino IDE, but it breaks functionality of Arduino-Makefile on Arch as of commit #b2d1ff23a2ff6d824aced575401e7b0deace672c.

sej7278:
this is because the arch arduino* packages aren't done correctly, sounds like they're setting AVR_TOOLS_DIR to /usr ?

AVR_TOOLS_DIR on debian for example is set to /usr/share/arduino/hardware/tools/avr/ and avrdude.conf is looked for in:

/usr/share/arduino/hardware/tools/avr/../
/usr/share/arduino/hardware/tools/avr/etc/

which then have a symlink inside: avrdude.conf -> /etc/avrdude.conf

so this matches the simply unzipped ide from arduino which has avrdude.conf in hardware/tools/avr/etc/avrdude.conf

you're welcome to make a PR that simply adds /etc/avrdude.conf to the searched locations, as you say that's where it will be on most distro's.

UltraBlackLinux:
I'd also like to have this fixed. Same problem currently

Fixed it with a symlink though

PR #588, opened 2018-10-23
Add a comment on the necessary config of AVRDUDE_CONF on some systems

Perdu:
As detailed in issue #576, some systems (e.g. Arch Linux) put avrdude.conf
outside of AVR_TOOLS_DIR. This can be fixed easily by manually setting the
path to look for in the AVRDUDE_CONF variable.


sej7278:
its already documented:

https://github.com/sudar/Arduino-Makefile/blob/master/arduino-mk-vars.md#avrdude-setting-variables

PR #597, opened 2019-01-11
Fix #594 and sam flag for ardmk-init

### LINKED ISSUE #594: Different Serial Port description when using Arduino IDE (Arduino micro)
Hi, I have run into an issue that is not critical, but somehow bothersome. When compiling/uploading a `.ino` source file into my Arduino Micro using the official IDE, the serial port description for the device reads `Arduino Micro`. But if I use the makefile offered here I get a serial port description that reads `USB IO Board`. **Note:** there is no problem with the uploaded program.

Here is how I test serial port descriptions:

    python -c 'from serial.tools.list_ports import comports; print([p.description for p in comports()])'

Does this mean I am missing some configuration steps? or is it an known (non-priority) issue?
is it possible for me to modify this string (it would be awesome)?

Thank you.

tuna-f1sh:
The fix for #594 to grab USB_PRODUCT and USB_MANUFACTURER from the boards.txt for Caterina (32u4) boards.

It also includes an additional flag (--sam) to `ardmk-init`, to configure for sam devices (include Sam.mk not Arduino.mk). I can separate if necessary but it seemed minor enough to include.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -716,6 +716,22 @@ ifeq ($(strip $(NO_CORE)),)
             endif
         endif
 
+        ifndef USB_PRODUCT
+            USB_PRODUCT := $(call PARSE_BOARD,$(BOARD_TAG),build.usb_product)
+            ifdef USB_PRODUCT
+                $(call show_config_variable,USB_PRODUCT,[COMPUTED])
+            endif
+        endif
+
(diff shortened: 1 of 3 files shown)


sej7278:
i noticed that - cheeky! its fine thanks

PR #599, opened 2019-02-25
Build the ArduinoCore API

szeder:
As part of the big modularizing efforts of the Arduino project they
split out the hardware-independent layer of the Arduino "language"
from the hardware-specific cores into the dedicated 'ArduinoCore-API'
repository.  As described in 'ArduinoCore-API's README, the API source
files won't reside directly in the directory of the standard Arduino
core, i.e. in 'ARDUINO_CORE_PATH', but in its 'ARDUINO_CORE_PATH/api'
subdirectory.  Consequently, Arduino-Makefile won't be able to build
any projects when using an Arduino core following the new directory
structure.

Prepare for the upcoming new Arduino core directory structure by
building all 'ARDUINO_CORE_PATH/api/*.cpp' source files as well.  Out
of caution, look out for and build any .c source files in that
directory, too: though there are no .c source files in the
'ArduinoCore-API' repository at the moment, in the future there might
be.

Note that a 'make clean' will be most likely necessary when
re-building a project after switching to the new directory structure.

Update: there was a bit of a grammar glitch in the commit message, hence the force-push.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -925,6 +925,10 @@ ifeq ($(strip $(NO_CORE)),)
         CORE_CPP_SRCS   = $(wildcard $(ARDUINO_CORE_PATH)/*.cpp)
         CORE_AS_SRCS    = $(wildcard $(ARDUINO_CORE_PATH)/*.S)
 
+        # ArduinoCore-API
+        CORE_C_SRCS    += $(wildcard $(ARDUINO_CORE_PATH)/api/*.c)
+        CORE_CPP_SRCS  += $(wildcard $(ARDUINO_CORE_PATH)/api/*.cpp)
+
         # USB Core if samd or sam
         ifeq ($(findstring sam, $(strip $(ARCHITECTURE))), sam)
             CORE_C_SRCS    += $(wildcard $(ARDUINO_CORE_PATH)/avr/*.c) # avr core emulation files
@@ -1159,7 +1163,7 @@ endif
 
 # Using += instead of =, so that CPPFLAGS can be set per sketch level
 CPPFLAGS      += -$(MCU_FLAG_NAME)=$(MCU) -DF_CPU=$(F_CPU) -DARDUINO=$(ARDUINO_VERSION) $(ARDUINO_ARCH_FLAG) \
-        -I$(ARDUINO_CORE_PATH) -I$(ARDUINO_VAR_PATH)/$(VARIANT) \
+        -I$(ARDUINO_CORE_PATH) -I$(ARDUINO_CORE_PATH)/api -I$(ARDUINO_VAR_PATH)/$(VARIANT) \
         $(SYS_INCLUDES) $(PLATFORM_INCLUDES) $(USER_INCLUDES) -Wall -ffunction-sections \
         -fdata-sections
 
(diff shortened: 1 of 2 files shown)


szeder:
I've considered introducing a new variable for the path of the ArduinoCore API, e.g. ARDUINO_API_PATH or something, but thought that we should wait for the first official Arduino release with this new directory structure before going that far.

In the meantime this patch is sufficient for building projects using cutting edge branches of various Arduino repositories.

sej7278:
seems i've just made a related PR, could you add the api directory to the include path (for e.g. HardwareSerial.h) like i have in #609 not just the c/cpp files, then we can probably merge this

szeder:
On Wed, Jul 03, 2019 at 05:45:13AM -0700, Simon John wrote:
> seems i've just made a related PR, could you add the api directory
> to the include path (for e.g. HardwareSerial.h) like i have in #609
> not just the c/cpp files, then we can probably merge this

Is it really necessary to add the `api` directory to the include
paths?

`Arduino.h` always start with `#include "api/ArduinoAPI.h"` in various
cores:

https://github.com/arduino/ArduinoCore-avr/blob/api/cores/arduino/Arduino.h#L23
https://github.com/arduino/ArduinoCore-megaavr/blob/namespaced_api/cores/arduino/Arduino.h#L23
https://github.com/arduino/ArduinoCore-samd/blob/api/cores/arduino/Arduino.h#L23

By explicitly specifying the `api/` directory prefix in that include
directive the compiler will be able to find that header file even if
that `api` directory is not added to the include path.
`api/ArduinoAPI.h` in turn then includes all other header files in
that directory with e.g. `#include "HardwareSerial.h"`:

https://github.com/arduino/ArduinoCore-API/blob/master/api/ArduinoAPI.h#L31

This means that the compiler will be able to find all those header
files as well,  because it starts looking for them in the directory
containing the file they are included from, even if that directory is
not added to the include path.

sej7278:
a quick "Blink" sketch with ```#include <HardwareSerial.h>``` at the top won't compile with your c/cpp bit added but without my include path bit added as it can't find the header file.

szeder:
On Thu, Jul 04, 2019 at 06:32:35AM -0700, Simon John wrote:
> a quick "Blink" sketch with ```#include <HardwareSerial.h>``` at the
> top won't compile with your c/cpp bit added but without my include
> path bit added as it can't find the header file.

To my understanding that ```#include <HardwareSerial.h>``` should not
be there at the top, the implicitly included `Arduino.h` takes care of
all that.

sej7278:
ok, yes it does seem to work if you ignore my include path bit and don't use the include - the binaries still have serial//uart/println functions in them

szeder:
On Thu, Jul 04, 2019 at 04:01:27PM +0200, SZEDER Gábor wrote:
> On Thu, Jul 04, 2019 at 06:32:35AM -0700, Simon John wrote:
> > a quick "Blink" sketch with ```#include <HardwareSerial.h>``` at the
> > top won't compile with your c/cpp bit added but without my include
> > path bit added as it can't find the header file.
> 
> To my understanding that ```#include <HardwareSerial.h>``` should not
> be there at the top, the implicitly included `Arduino.h` takes care of
> all that.

Hang on for a sec.  Even though this `#include <HardwareSerial.h>` is
unnecessary, it's not a good excuse to fail the build if it is
present.

OK, I'll update soon-ish.

szeder:
Done, `$(ARDUINO_CORE_PATH)/api` is now added to the list of directories to be searched for header files, and the commit message is updated accordingly.

sej7278:
thanks for the PR, i've merged now

PR #603, opened 2019-04-01
add swtich for ARM 'avr-size' so that .elf is passed as arg

tuna-f1sh:
`avr-size` was defaulting to AVR non-aware when using ARM builds, which meant passing .hex rather than .elf to size and limiting size readout.

Adding a switch for defining `avr-size` and SIZEFLAGS to Sam.mk seems like the best way to include proper ARM support (despite counter-intuitive naming).

I also corrected the GNU extensions flag to gnu11 to match the Arduino arm support as it was now failing to compile wiring.c


sej7278:
That's for the PR @tuna-f1sh

Issue #601, opened 2019-03-20
PARSE_BOARD chops off flags which contain '=' signs


beldenfox:
Teensy.mk calls on PARSE_BOARD to pick up compiler flags verbatim from boards.txt. If the line contains an option of the form '-flag=blah' the line will be chopped off at the equal sign. For example:

`teensy31.build.flags.cpu=-mthumb -mcpu=cortex-m4 -fsingle-precision-constant`

will be parsed as

`teensy31.build.flags.cpu=-mthumb -mcpu`

The problem is in the  first `cut` command in the macro which should pick up everything from the first equal sign to the end of the line. So instead of

`cut -d = -f 2`

it should read

`cut -d = -f 2-`

szeder:
I believe you are right and the parser should indeed include everything following the first '=' up to the end of the line.

This used to work for Teensy, but regressed in 2442daf (Moved the PARSE_BOARD macro from Arduino.mk to Common.mk., 2018-09-13) in PR #580.

(On a slightly related note: [`Sam.mk` still has its own `PARSE_BOARD` macro](https://github.com/sudar/Arduino-Makefile/blob/master/Sam.mk#L135))

@wingunder

tuna-f1sh:
Seems like it would be a good idea to fix this in Common.mk and remove the `IFNDEF PARSE_BOARD...` in Sam.mk since it is now in Common.mk. The `IFNDEF` should mean it is not overwritten in Sam.mk but removing it will make it cleaner/easier to maintain.

stv0g:
I created a PR for the proposed solution.

I also checked that `PARSE_BOARD` from `Sam.mk` and `Common.mk` are equivalent.

PR #604, opened 2019-05-03
Fix regression: PARSE_BOARD chops off flags which contain '=' signs

stv0g:
See #601


sudar:
@sej7278 

I checked the diff and it looks good to me, but since I don't have a non-Arduino board I am not sure if it causes any other issues.

stv0g:
Hi @sej7278 @sudar 

any updates on this PR?

Cheers,
Steffen

sej7278:
doesn't look like it'll cause problems and would be nice to remove duplicated code from sam.mk :+1:

sudar:
@stv0g 

PR #605, opened 2019-06-12
Show the configuration when ARDUINO_QUIET=0

szeder:
There is a bit of inconsistency between documentation and code
regarding the ARDUINO_QUIET variable: 'arduino-mk-vars.md' states
that ARDUINO_QUIET "Defaults to `0` (unset/disabled)", but the code
only checks whether it's defined or not, and doesn't check whether
it's set to '0' or something else.

Consequently, having 'ARDUINO_QUIET=0' in the Makefile or running
'make ARDUINO_QUIET=0' contadicts the documentation and doesn't print
the configuration.  It also means that if someone in general prefers
not to see a screenful of configuration on each build and therefore
has 'ARDUINO_QUIET = 1' in the project's Makefile or 'config.mak',
then there is no way to override it from the command line in the odd
case when showing the configuration is desired.

Modify the corresponding condition in Arduino.mk to check whether
ARDUINO_QUIET is set to 0 and treat an undefined ARDUINO_QUIET
variable as "set to 0" as well.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -241,6 +241,9 @@ arduino_output =
 # running for the first time (i.e., not after a restart after
 # regenerating the dependency file), then output the configuration.
 ifndef ARDUINO_QUIET
+    ARDUINO_QUIET = 0
+endif
+ifeq ($(ARDUINO_QUIET),0)
     ifeq ($(MAKE_RESTARTS),)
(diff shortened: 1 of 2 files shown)


sej7278:
merged thanks @szeder 

yes i noticed this myself before and meant to fix it but forgot about it!

PR #608, opened 2019-07-02
Allow custom link script (Teensy)

mumme74:
This change makes it possible to set a custom linker script.


sudar:
Thanks @mumme74 

Just merged it.

PR #609, opened 2019-07-03
Added api subdirectory to $ARDUINO_CORE_PATH for new cores like megaavr

sej7278:
seems to work fine even with cores where there is no api subdirectory


sej7278:
edit: don't merge, seems we have a duplicate with #599

sej7278:
no longer needed

PR #612, opened 2019-07-13
Fix renaming PRODUCT and MANUFACTURER for Caterina devices like leonardo

G33kLabs:
It works with that in the `Makefile` (all without quotes) :

```
USB_MANUFACTURER = MANUFACTURER_NAME
USB_PRODUCT = PRODUCT_NAME
```


sej7278:
this seems to break cores that actually use these settings e.g. the adafruit ones, where they use quotes:

```
feather32u4.build.usb_product="Feather 32u4"
feather32u4.build.usb_manufacturer="Adafruit"
```

the makefile without your patch works fine if you also use quotes, which you should do anyway as the strings may have spaces:

```USB_PRODUCT = "product with spaces"```

G33kLabs:
Ok, will check you that tomorrow with electronics near.

sudar:
@G33kLabs 

Did you get a chance to check it with the real hardware?

PR #614, opened 2019-07-31
Rename `CTAGS_CMD` to `CTAGS_EXEC` (in the documentation).

co-dan:
The former is not actually tweakable (anymore)

### DIFF
--- a/arduino-mk-vars.md
+++ b/arduino-mk-vars.md
@@ -1878,16 +1878,16 @@ CTAGS_OPTS = -V
 
 ----
 
-### CTAGS_CMD
+### CTAGS_EXEC
 
 **Description:**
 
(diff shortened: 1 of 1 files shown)


co-dan:
I think the CI fails because of the way it's set up atm 

```
Cannot uninstall 'pyserial'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
You are using pip version 19.0.3, however version 19.2.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
The command "tests/script/bootstrap.sh" failed and exited with 1 during .
```

sudar (approved):

sudar:
Thanks @co-dan for the PR. Just merged it.

PR #617, opened 2019-11-08
Remove requirement for library.properties file for user libraries.

AlfTetzlaff:
I removed the requirement for user libraries to provide a library.properties file, which caused quite a headache for me, because I wondered why my own written libraries were never found.

One could further alter the provided solution in order to allow for separate /include and /src subfolders, which is also a common file structure.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -63,7 +63,7 @@
 #
 # On Windows declare this environmental variables using the windows
 # configuration options or Cygwin .bashrc. Control Panel > System > Advanced system settings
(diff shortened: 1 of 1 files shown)


sudar:
@AlfTetzlaff Thanks for the PR.

@sej7278 I like the idea of using different directories for the user code and libraries. But I am not sure if it will break any of our existing assumptions. What are your thoughts on this?

sej7278:
i don't think this will break things and the missing library.properties thing has bitten me recently for my own libs :+1:

PR #618, opened 2019-11-13
Test fixes

alissa-huskey:
Various fixes to test scripts:

bootstrap.sh
- Update the download link. Also, check if the downloaded file is a zip file and print a sensible error message if not.

runtests.sh
- Set the directory that the script can be run from anywhere
- Add a -q option for quiet mode

Common.mk
- Fix the DEPENDENCIES_ARDUINO_DIR for mac
- Instead of hardcoding the DEPENDENCIES_MPIDE_DIR, find the newest dir matching mpide-0023-* in DEPENDENCIES_ARDUINO_DIR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 build-cli
 /.project
 build-*
+.tags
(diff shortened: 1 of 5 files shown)


sudar:
@alissa-huskey 

Thanks for the PR.

@sej7278 

What are your thoughts? I like the `info` helper.

sej7278:
no idea @sudar, travis seems to break more often than its worth, so anything that attempts to fix it is great!

ladislas:
what about moving to github actions? they offer much more than travis, are free and you can also run macOS to test this platform, not just linux.

alissa-huskey:
Just noticed this was still open. Thanks for incorporating my work!

PR #619, opened 2019-11-13
Add PLATFORM_PATH

alissa-huskey:
Add PLATFORM_PATH variable to the root hardware directory for a given
architecture to avoid duplication between boards.txt, cores, platform
libraries, etc.

I needed to access this value and was hoping to inherit it, when I learned that it did not exist as a distinct variable in Arduino.mk. It seemed like it would be generally useful and a bit more DRY to define it independently. So I thought I would offer it in case you find it useful. No worries if you'd rather not.

I opted to avoid setting PLATFORM_PATH in the case where any of the ALTERNATE_CORE variables are set. I realize that there is some duplication between the concepts and it may be confusing. I figured sticking with the minimally invasive simple refactoring was safest.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -63,7 +63,7 @@
 #
 # On Windows declare this environmental variables using the windows
 # configuration options or Cygwin .bashrc. Control Panel > System > Advanced system settings
-# The paths must use Unix style forward slash and not have any spaces 
+# The paths must use Unix style forward slash and not have any spaces
 # or escape charactors. One must use a symbolic link if the path does
 # contain spaces.
 #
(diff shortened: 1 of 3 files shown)


sudar:
@alissa-huskey 

Thanks for your contribution and sorry for the delay in getting back to you.

@sej7278 

sej7278:
not really sure what the use case for this would be if its not used with `ALTERNATE_CORE`, but it shouldn't hurt as it defaults to the original value if not set.

alissa-huskey:
Just noticed this is still open.

It's been awhile since I've worked on my Arduino projects, so I don't recall my original use-case. I'll close this and if I run into it again I'll resubmit with more information.

PR #620, opened 2019-11-14
Python3 migration

### LINKED ISSUE #542: Eventual Python3 migration
```/usr/bin/env python``` still points to 2.7 for me on Debian Sid, so its far from a problem for now, but python2.7 is EOL in 2 years.

i'm just thinking ard-reset-arduino, robotis-loader and ardmk-init should probably change their shebangs to use ```/usr/bin/python3``` instead.

Also the arduino-mk.spec and debian/control files will need updating to use python3-serial instead of python-serial (debian) and python3-pyserial instead of pyserial (fedora).

Need to check MacOS/cygwin implications and run by @maqifrnswa and @tuna-f1sh too.

https://www.debian.org/doc/packaging-manuals/python-policy/ch-python3.html

sej7278:
Fixes #542

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -861,7 +861,7 @@ ifndef RESET_CMD
   endif
   ifneq (,$(findstring CYGWIN,$(shell uname -s)))
       # confirm user is using default cygwin unix Python (which uses ttySx) and not Windows Python (which uses COMx)
-      ifeq ($(shell which python),/usr/bin/python)
+      ifeq ($(shell which python3),/usr/bin/python3)
         RESET_CMD = $(ARD_RESET_ARDUINO) $(ARD_RESET_OPTS) $(DEVICE_PATH)
       else
         RESET_CMD = $(ARD_RESET_ARDUINO) $(ARD_RESET_OPTS) $(call get_monitor_port)
--- a/README.md
+++ b/README.md
@@ -83,14 +83,14 @@ installer or download the distribution zip file and extract it.
 The Makefile also delegates resetting the board to a short Python program.
 You'll need to install [`pySerial`](https://pypi.python.org/pypi/pyserial) to use it though.
 
-On most systems you should be able to install it using either `pip` or `easy_install`.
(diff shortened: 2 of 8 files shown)


tuna-f1sh:
Following the discussions in #542 regarding not using macOS system Python, I've done what I suggested and added a `python` resolution to Common.mk. It is then pre-appended to the `ard-reset-arduino` scripts.

https://github.com/tuna-f1sh/Arduino-Makefile/commit/87d524107147af571abf2a1515e98f5422c5f844

I've merged your changes for Python 3 support too so looks like we finally have fix?!

https://github.com/tuna-f1sh/Arduino-Makefile/commit/3241c40b1b3199060036babc8f8d88f4efc5408b

sej7278:
yeah i quite like the PYTHON_CMD, as the way its going we're likely to need to set it for macos, i'm not entirely keen on basically ignoring the shebang on linux, but at this point i've given up a bit!

i was thinking alternatively to just changing the shebang as part of debian/fedora packaging and defaulting to "/usr/bin/env python"

tuna-f1sh:
I considered this too. I guess the problem with this approach is people using the repository clone.

happyaron:
If I understand correctly, `/usr/bin/env python` will eventually return python3 when the migration is complete at the distro's PoV, so this imposes the program supports both python2/3 and would like to use distro's default.

tuna-f1sh:
@happyaron if only it were that simple: see https://github.com/sudar/Arduino-Makefile/issues/542#issuecomment-535661160; `/usr/bin/env python` is not allowed in rpm/deb packages, they must be absolute.

sej7278:
so if we use `/usr/bin/env python` and then just patch it when we build rpm/deb packages, will that work with the various macos implementations?

it'll still work on linux without hardcoding the shebang (git checkout) too.

tuna-f1sh:
Yes, I think that is actually the simplist/best solution.

sej7278:
ok done - reset to using `env python` by default and added a sed to the rpm specfile.

will have to think about the debian patch but that's outside of this repo anyway.

PR #630, opened 2020-04-12
Let's make examples platform neutral

osamuaoki:
Depending on platform,  ARDMK_DIR can be:
 * `../..`
* `/usr/local`
* `/usr/share/arduino/`
* `/home/username/Arduino-Makefile`
* ...

And `README.md` tells people to set up ARDMK_DIR environment variable to ease platform differences, we should follow it in examples. 

Here is my humble try.

This helps Debian bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=949769

### DIFF
--- a/examples/ATtinyBlink/Makefile
+++ b/examples/ATtinyBlink/Makefile
@@ -54,6 +54,6 @@ BOARD_TAG = attiny85
 # ------------------------------------------------------------------ #
 
 # Path to the Arduino Makefile
-include /usr/share/arduino/Arduino.mk
+include $(ARDMK_DIR)/Arduino.mk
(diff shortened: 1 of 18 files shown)


sej7278:
yes i don't like the `include ../../Arduino.mk` that's in most of the examples, that's really for git-checked out users.

i like your idea of changing that to `include $(ARDMK_DIR)/Arduino.mk` as it will then default to "relative to Common.mk" which will be /usr/share/arduino on debian, but will also work for git-checkouts too.

:+1:

sudar:
Thanks @osamuaoki for the PR.

PR #640, opened 2020-08-06
Fix Travis CI, add SAMD test support, support GNU grep on macOS

### LINKED ISSUE #625: Multiple toolchains cause makefile failure
It appears that if more than one directories exist in the below directory, it causes ARM_TOOLS_DIR to contain multiple space separated paths and the makefile to fail

Sam.mk:165 `$(ARDUINO_PACKAGE_DIR)/$(ARDMK_VENDOR)/tools/$(TOOL_PREFIX)-gcc`
https://github.com/sudar/Arduino-Makefile/blob/e870443f4824cbbcd560f500c6072012dd668f62/Sam.mk#L165

I discovered this as my directory reference contained: "4.8.3-2014q1" and "7-2017q4"

tuna-f1sh:
Content as subject, includes fix #625. Includes @sej7278's Python 3 branch. BSD `grep` was creating some warnings parsing boards.txt that were annoying!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-sudo: required
+os: linux
+dist: xenial
 language: c
 compiler:
   - gcc
@@ -9,3 +10,6 @@ addons:
     packages:
       - "python3"
       - "python3-pip"
+env:
+  global:
+    - ARDMK_DIR=$TRAVIS_BUILD_DIR
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,8 @@ compiler:
   - gcc
 script: tests/script/runtests.sh
 before_install: tests/script/bootstrap.sh
+addons:
+  apt:
+    packages:
+      - "python3"
+      - "python3-pip"
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1227,15 +1227,24 @@ CFLAGS        += $(CFLAGS_STD)
 CXXFLAGS      += -fpermissive -fno-exceptions $(CXXFLAGS_STD)
 ASFLAGS       += -x assembler-with-cpp
 DIAGNOSTICS_COLOR_WHEN ?= always
-ifeq ($(shell expr $(CC_VERNUM) '>' 490), 1)
-    ASFLAGS  += -flto
-    CXXFLAGS += -fno-threadsafe-statics -flto -fno-devirtualize -fdiagnostics-color=$(DIAGNOSTICS_COLOR_WHEN)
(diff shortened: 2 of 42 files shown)


tuna-f1sh:
Now includes https://github.com/sudar/Arduino-Makefile/pull/630 (manual as @osamuaoki's branch has gone awol).

sej7278:
:+1:

sej7278:
> Now includes #630 (manual as @osamuaoki's branch has gone awol).

i'm still not 100% sure about that one.

in the debian package it doesn't work as `ARDMK_DIR` isn't defined anywhere and even if you define it, you can't build the examples as you can't write to `/usr/share/doc/arduino-mk/examples`

at least with `../../Arduino.mk` it works for a git checkout.

seems like the debian package would require the user to copy the examples to their sketchbook directory and define `ARDMK_DIR` in their local Makefile, like:

```
ARDMK_DIR = /usr/share/arduino
#ARDMK_DIR = ../../
BOARD_TAG = uno
include $(ARDMK_DIR)/Arduino.mk
```

at which point setting `ARDMK_DIR` seems a bit pointless and you may as well stick with:

```
BOARD_TAG = uno
include /usr/share/arduino/Arduino.mk
```

so maybe we should drop that PR?

tuna-f1sh:
I thought the thinking was that since the README mandates that `ARDMK_DIR` should be defined by the user, the examples failing helps to ensure this?

Could add in a catch to check for `ARDMK_DIR` and error if undefined with helpful hint or define it to as it was prior:

```
ifdef ARDMK_DIR
    include $(ARDMK_DIR)/Arduino.mk
else
    $(info ! Using project relative ARDMK_DIR, please define environment variable ARDMK_DIR as Arduino-Makefile installation directory for external use)
    include ../../Arduino.mk
endif
```

sej7278:
where would you put that though - in every example?

tuna-f1sh:
Yes would be quite messy. It does make the example Makefiles better for copying into a project however. 

Don't mind, can just remove it and revert back to relative path for all the Makefiles.

sej7278:
yeah if you can revert it back to ../../ we can look at it again afterwards maybe

tuna-f1sh:
It's gone.

tuna-f1sh:
Have you had chance to review @sudar ? Would be good to get a 2.0 release after this push to fix the Python 3 shebang problem.

sudar:
@tuna-f1sh 

I finally got some time this weekend to test it.

I will merge it mostly by tomorrow.

sudar:
Sorry guys, what I thought will take just 2 days ended up being 2 weeks. Finally I was able to test the fix and have just merged it.

Thanks @sej7278 @tuna-f1sh and others who helped to get this done.

PR #645, opened 2020-10-19
Use C++ compiler when linking for SAMD-based boards

### LINKED ISSUE #644: ELF link uses $(CC), not $(CXX)
I've just started using this Makefile kit (it's quite nice) for a SAMD-based small project: the project currently contains one .ino file and some .hpp files (which are '#include'-d into the .ino file for the moment).

I got everything to compile, but the final link to produce the ELF file failed with a number of missing symbols, all of which come from the C++ standard library. While looking at Arduino.mk [here](https://github.com/sudar/Arduino-Makefile/blob/c7eb909b849e354fdb7630d11405ea6de706f1ba/Arduino.mk#L1679) I noticed that the linking is being done by calling $(CC) instead of $(CXX), which means the C++ standard library doesn't get included into the link automatically.

While I've made this change in my local Arduino.mk and now the project links successfully, I can't help but think that this is something which shouldn't be necessary, as pretty much all Arduino projects (at least those which originated in the IDE) are in C++ and require the standard library. Is there some sort of configuration setting I've missed?

kpfleming:
The Arduino IDE supports full usage of C++ (including the standard
library) on SAMD-based boards, so the final linking of the executable
must be done using the C++ compiler.

Closes #644.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1674,9 +1674,9 @@ pre-build:
 
 # copied from arduino with start-group, end-group
 $(TARGET_ELF): 	$(LOCAL_OBJS) $(CORE_LIB) $(OTHER_OBJS)
-# sam devices need start and end group
+# sam devices need start and end group, and must be linked using C++ compiler
 ifeq ($(findstring sam, $(strip $(ARCHITECTURE))), sam)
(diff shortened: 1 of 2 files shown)


sej7278:
merged thanks @kpfleming

PR #646, opened 2020-10-20
Improve detection of ard-reset-arduino location

kpfleming:
Find location of ard-reset-arduino before prefixing it with PYTHON_CMD (so that 'which' failure can be detected). Also display the resulting RESET_CMD, or user-provided RESET_CMD.

Prior to this patch the `ifndef ARD_RESET_ARDUINO` would never be triggered because ARD_RESET_ARDUINO contained the value of PYTHON_CMD, even if the 'which' operation failed.

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -863,12 +863,13 @@ endif
 # Reset
 
 ifndef RESET_CMD
-  ARD_RESET_ARDUINO := $(PYTHON_CMD) $(shell which ard-reset-arduino 2> /dev/null)
-  ifndef ARD_RESET_ARDUINO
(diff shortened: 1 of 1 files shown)


sej7278:
yup, got you - it'll always be defined as it contains `PYTHON_CMD`, good spot, merged thanks

kpfleming:
Thanks! I couldn't figure out why 'make upload' always had an empty reset command :-)

With this patch merged and the other two you merged over the past day, I now have an MKR1010WiFi project working perfectly with Arduino-Makefile... build, upload, and monitor. The only remaining things I need the IDE for are updating the toolchains and libraries, and OTA uploads.

sej7278:
oh good, i was going to ask if the merges had fixed your wifi build :+1:

PR #659, opened 2021-04-21
Add newest ard-reset-arduino flag support.

JDuchniewicz:
The option `--zero` no longer works with Arduino 1.8.13 for SAMD boards and has to be replaced with the `--caterina` option.


tuna-f1sh:
Please can you add some more context to ‘no longer works’?

I’m still using `—zero` with the Arduino SAMD bootloader with no issue.

JDuchniewicz:
The `ard-reset-arduino` does not have the `--zero` option anymore in my case (1.6.0 version of the program).

tuna-f1sh:
The `--zero` option was added with the SAMD support - not removed.

The release version of this repository (1.6.0) does not include support for SAMD. You need install `ard-reset-arduino` from the master/HEAD and not use a package manager.

JDuchniewicz:
Hmm this is embarrassing. I thought that the upstream changes removed that option... Closing it in this case 😄

PR #678, opened 2022-10-04
Fix: 'grep: warning: stray \ before #' in GNU grep >= 3.8

jenda122:
After upgrading GNU grep in Debian from 3.6 to 3.8, I get lots of "grep: warning: stray \ before #" warnings. This fixes it.


liam-clink:
I was just about to submit just the same pull request. The reason for this btw is that it's bringing grep into line with the POSIX standard. See: https://github.com/koalaman/shellcheck/issues/2573

PR #686, opened 2026-02-20
Fix format flag for avr-size command

badrihippo:
I was unable to compile programs because the `avr-size` command failed saying that "C is an unrecognised option".

It seems like `avr-size` no longer recognises the `-C` option; you have to pass in the full form of `--format=avr` instead. My workaround for now is to add the following line to my `Makefile`:

```makefile
SIZEFLAGS     ?= --mcu=$(MCU) --format=avr
```

However, updating the line here should fix it for everyone (and if I'm not wrong, it'll even be backwards-compatible with older versions of `avr-size` that would've still recognised the old `-C` option!)

### DIFF
--- a/Arduino.mk
+++ b/Arduino.mk
@@ -1260,7 +1260,7 @@ else
 endif
 
 LDFLAGS       += -$(MCU_FLAG_NAME)=$(MCU) -Wl,--gc-sections -O$(OPTIMIZATION_LEVEL)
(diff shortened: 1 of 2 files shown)


pupitetris:
Please accept this pull request. I also just bumped into this error after installing arduino-mk on a Debian 13.3 (Trixie) machine.

